Issue #11, opened 2018-02-24
Reverse proxy for non-swarm containers


ibnesayeed:
While the Swarm mode is the way to deploy services and it might become the default at some point, many people are still deploying individual containers directly. Theoretically, the plugin should work in those situations as well, but we need to test and document it.

pwFoo:
Would be nice to use with native docker or rancher (2.x is Kubernetes based) too. :)

lucaslorentz:
@pwFoo For Rancher + Kubernetes I suggest looking at https://github.com/wehco/caddy-ingress-controller

PR #13, opened 2018-02-25
Add support for proxying containers

lucaslorentz:
Closes https://github.com/lucaslorentz/caddy-docker-proxy/issues/11

Changes:
* Add support for proxying directly to containers, without swarm services

@ibnesayeed Can you please take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,27 @@
 # CADDY-DOCKER-PROXY [![Build Status](https://travis-ci.org/lucaslorentz/caddy-docker-proxy.svg?branch=master)](https://travis-ci.org/lucaslorentz/caddy-docker-proxy)
 
 ## Introduction
-This plugin enables caddy to be used as a reverse proxy for Docker Swarm services.
+This plugin enables caddy to be used as a reverse proxy for Docker.
 
 ## How does it work?
-It scans Docker Swarm services metadata looking for labels indicating that the service should be exposed on caddy.
+It scans Docker metadata looking for labels indicating that the service or container should be exposed on caddy.
 
-Then it generates an in memory Caddyfile with website entries and proxies directives pointing to each Docker Service internal DNS name.
+Then it generates an in memory Caddyfile with website entries and proxies directives pointing to each Docker service dns name or container IP.
 
-Every time Docker services changes, it updates the Caddyfile and triggers a caddy zero-downtime reload.
+Every time Docker objects changes, it updates the Caddyfile and triggers a caddy zero-downtime reload.
 
-## Basic service labels
-To expose a service inside caddy configuration, you just need to add labels starting with caddy to the service.
+## Basic labels
+To expose a service or container inside caddy configuration, you just need to add labels starting with caddy.
 
 Those are the main labels that configures the basic behavior of the proxying:
 
 | Label | Example | Description |
 | - | - | - |
-| caddy.address | service.example.com | addresses that should be proxied to that service separated by ',' |
-| caddy.targetport | 80 | the port being serverd by the service |
-| caddy.targetpath | /api | the path being served by the service |
+| caddy.address | service.example.com | addresses that should be proxied separated by ',' |
+| caddy.targetport | 80 | the port being server by container |
+| caddy.targetpath | /api | the path being served by container |
 
-The values above will generate the following caddy configuration:
+When added to a service, the values above will generate the following caddy configuration:
 ```
 service.example.com {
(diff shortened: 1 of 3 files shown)


lucaslorentz:
I wonder if we need to support proxying same caddy website to multiple containers.

Something like:
* Group containers by caddy.address.
* For each group:
  - generate caddy config using labels from first container
  - generate proxy directive targetting all containers IPs of that group

README.md
```diff
@@ -1,27 +1,27 @@
...
 ## Introduction
-This plugin enables caddy to be used as a reverse proxy for Docker Swarm services.
+This plugin enables caddy to be used as a reverse proxy for Docker.
 
 ## How does it work?
-It scans Docker Swarm services metadata looking for labels indicating that the service should be exposed on caddy.
+It scans Docker metadata looking for labels indicating that the service or container should be exposed on caddy.
 
-Then it generates an in memory Caddyfile with website entries and proxies directives pointing to each Docker Service internal DNS name.
+Then it generates an in memory Caddyfile with website entries and proxies directives pointing to each Docker service dns name or container IP.
```
ibnesayeed: dns => DNS

README.md
```diff
@@ -1,27 +1,27 @@
...
 
-Every time Docker services changes, it updates the Caddyfile and triggers a caddy zero-downtime reload.
+Every time Docker objects changes, it updates the Caddyfile and triggers a caddy zero-downtime reload.
```
ibnesayeed: Docker objects => a Docker object

ibnesayeed:
> I wonder if we need to support proxying same caddy website to multiple containers.

That is not a bad idea. In case of Swarm mode, replicated services are already covered, but bare containers can also benefit from it. However, grouping by the dress is not be safe. We will have to use all three (or perhaps all) `caddy.*` labels combined for grouping.

lucaslorentz:
Good.
I'm not sure if many users would have replicated bare containers on the same host.
We can tackle that in another PR/issue.

pwFoo:
What's about docker compose services without swarm? Multiple backend container should be loadbalanced for the same public domain.

ibnesayeed:
> @lucaslorentz: I'm not sure if many users would have replicated bare containers on the same host.

As per the following, it is very likely to have multiple replicas because Compose makes it easy to scale services and it's been there long before Swarm mode was added to the Docker core.

> @pwFoo: What's about docker compose services without swarm? Multiple backend container should be loadbalanced for the same public domain.

lucaslorentz:
@pwFoo 

I think Docker Swarm Stacks is the new docker-compose.

Even for isolated machines, Docker Swarm would be beneficial because it ensures your containers are always running. As far as I know docker-compose only creates/destroy containers when you call the command, it doesn't monitors in order to ensure the desired number of replicas are running.

I would like to understand better your scenario in order to do the right features. Can you please describe it a bit further?

- Do you want to run caddy inside a container or outside?
- How many docker hosts caddy needs to interact with?
- Do you have multiple replicas for same application? Or are you talking about multiple applications sharing the same public domain under different paths?

ibnesayeed:
> Even for isolated machines, Docker Swarm would be beneficial because it ensures your containers are always running.

While this is very true, I think the Swarm mode was a little late entry in the ecosystem and by that time many people were already using Compose. Docker is still actively maintaining Compose separately because the `docker stack` is not a complete replacement of the `docker-compose` command yet. There are many things desired from it if the core docker commands were to eliminate the need of compose, especially in the development mode. I think docker is working in that direction, but GitHub is full of various projects that come with the `docker-compose.yml` file and instructions on how to get started with compose.

That said, I think if we start grouping containers for proxying based on `${caddy.address}:${caddy.targetport}/${caddy.targetpath}` then it won't matter how they were created.

lucaslorentz:
Makes sense, I will do a new PR.

pwFoo:
Hello @lucaslorentz ,

will loadbalancing added soon?
At the moment If I increase replicas each request is routed to the first container / caddyfile entry.
```
http://192.168.199.22/httpd1 {
  proxy / 10.0.2.5:80
}
http://192.168.199.22/httpd1 {
  proxy / 10.0.2.6:80
}
```

It should be balanced to the backend containers:
```
http://192.168.199.22/httpd1 {
  proxy / 10.0.2.5:80 10.0.2.6:80
}
```

lucaslorentz:
@pwFoo can you please create an issue with that?
Just as a reminder and we can track it there.

PR #17, opened 2018-03-06
Delay loader initialization until caddy calls it

lucaslorentz:
Loader was initialized before caddy invokes it, that was causing 2 problems:
- When run unit tests it was connecting to docker and printing to console
- It was always logging first caddyfile content to stdout, not respecting `-log file` caddy argument

@ibnesayeed


ibnesayeed:
LGTM!

Issue #15, opened 2018-03-03
Negotiate API version instead of relying on DOCKER_API_VERSION alone


jakebailey:
Using `DOCKER_API_VERSION` works, however the Docker client API is also able to negotiate an API version itself:

```go
dockerPing, err := dockerClient.Ping(ctx)
// deal with err
dockerClient.NegotiateAPIVersionPing(dockerPing)

// or

dockerClient.NegotiateAPIVersion(ctx)
```

The ping response includes the Docker daemon's max supported API version, and the client will set the right options to make everything work. You can also call `NegotiateAPIVersion`, which does the same thing, but discards the errors and uses defaults if something goes wrong. The former's probably better, since you can bail out early if `Ping` fails (i.e. you can report a problem with the connection to Docker before trying to start using it).

It'd be a good idea to stick this into `GenerateCaddyFile` after you create the client.

lucaslorentz:
Sounds good, it would be much better to don't have to set DOCKER_API_VERSION.

A PR is very welcome 😄

jakebailey:
It appears I can't do this (at least not when I work within vendored deps), because the version in your glide lock is very out-of-date and doesn't include any of the negotiation functions. API negotiation was added the Docker client 9 months ago, and the version you have pinned doesn't include it.

Unfortunately, after the whole moby thing, the main `docker/docker` repo stopped getting branches/tags/releases, so you'd have to pin to a revision. The Docker CLI client (`docker/cli`) uses `0ede01237c9ab871f1b8db0364427407f3e46541`, but I personally just work against master for my projects.

PR #18, opened 2018-03-07
Update dependencies + Add Docker Api Version Negotiation

lucaslorentz:
@jakebailey Docker Api Version Negotiation works very well!

@jakebailey @ibnesayeed Can you please review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,6 @@ You can modify docker connection using the following environment variables:
 * **DOCKER_CERT_PATH**: to load the tls certificates from.
 * **DOCKER_TLS_VERIFY**: to enable or disable TLS verification, off by default.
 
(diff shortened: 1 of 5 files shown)


jakebailey:
I suggest doing the following instead:
```go
dockerPing, err := dockerClient.Ping(context.Background())
if err != nil {
	addError(&buffer, err)
	return buffer.Bytes()
}

dockerClient.NegotiateAPIVersionPing(dockerPing)
```
That way, you'll return an error whenever the client hasn't connected to docker, rather than failing when using the client in the future.

ibnesayeed:
@jakebailey's suggestion above seems very reasonable.

lucaslorentz:
Fixes issue: https://github.com/lucaslorentz/caddy-docker-proxy/issues/15

PR #19, opened 2018-03-12
Listen to docker events and allow fallback to other caddy loaders

lucaslorentz:
This PR changes caddy-docker-proxy to listen to docker events in order to update caddy almost immediately: 100ms delay.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ caddy
 caddy.exe
 vendor
 debug.test
+Caddyfile
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -15,28 +15,11 @@ import (
 	"github.com/docker/docker/client"
 )
 
-var dockerClient *client.Client
 var suffixRegex = regexp.MustCompile("_\\d+$")
 
 // GenerateCaddyFile generates a caddy file config from docker swarm
-func GenerateCaddyFile() []byte {
+func GenerateCaddyFile(dockerClient *client.Client) []byte {
 	var buffer bytes.Buffer
-
-	if dockerClient == nil {
-		var err error
-		if dockerClient, err = client.NewEnvClient(); err != nil {
-			addError(&buffer, err)
-			return buffer.Bytes()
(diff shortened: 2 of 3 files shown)


lucaslorentz:
@ibnesayeed Can you please take a look?

lucaslorentz:
The delay is intentional, to throttle events.

ibnesayeed:
What is the purpose of calling `update()` in `reloadIfChanged` is false? Also, since `AfterFunc()` deferred call happens in a separate thread, we should check carefully if there is any possibility of a race condition when calling updates or setting/resetting timers. I am also not quite clear about the throttling logic, because if we just delay every event by a set amount of time, their relative execution delay will not change and if we keep pushing a shared timer back for successive events before they expire then there will be starvation like situation.

The code is not necessarily bad, but I just wanted to pick your brain on the logic you used and how you envision it to function.

lucaslorentz:
The first update fills the input content for the first time, there is no need to reload caddy on that case because it is still loading the configuration for the first time.

I've positioned changes to skipEvents and timer reset in a way I didn't envision problems with race conditions. The only race condition I see is if an event arrives between the execution of line 106 and 108, and as result it will unnecessarily check for changes in docker again 100ms later. I don't see any harm on that.

The starvation in case of successive events is a good point, that's why I set `skipEvents = true` right before pushing the timer back. So next events will not push the timer back again. Guaranteeing the throttle behavior, instead of debounce.

Let me know if you have some other concerns.

lucaslorentz:
I'm not sure about relying only on events, that's way I didn't remove the pooling that happens every 10 seconds.

I also tested listening to events in docker swarm and without, it works fine on both cases.

Another change is that now if it can't connect to docker, the caddyfile will be empty, and caddy will fallback to the next caddyfile loader.

That would allow you to use a caddy binary with docker plugin even when you don't wan't to connect to docker.

ibnesayeed:
Thanks, it sounds good to me and the code LGTM!

Issue #28, opened 2018-04-30
Wrong swarm network ip address


pwFoo:
Hi @lucaslorentz,

I have a problem with the plugin.

It uses the wrong ip address and can't reach the backend container.

```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet 10.0.4.7/32 brd 10.0.4.7 scope global lo
       valid_lft forever preferred_lft forever
    inet 10.0.3.11/32 brd 10.0.3.11 scope global lo
       valid_lft forever preferred_lft forever
259: eth0@if260: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP 
    link/ether 02:42:0a:00:04:0a brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.0.4.10/24 brd 10.0.4.255 scope global eth0
       valid_lft forever preferred_lft forever
```

caddyfile
```
example.com {
  errors stdout
  log stdout
  proxy / 10.0.4.10:2015
}
```

Ping test from web proxy container.
```
docker exec -ti proxy_webproxy.1.zr2bmrav82ok717yu7j51jthp ping 10.0.4.10
PING 10.0.4.10 (10.0.4.10): 56 data bytes
^C
--- 10.0.4.10 ping statistics ---
4 packets transmitted, 0 packets received, 100% packet loss
```

Added an alias to networks section to get a hostname. It's resolved to `10.0.3.11`
```
docker exec -ti proxy_webproxy.1.zr2bmrav82ok717yu7j51jthp ping webspace1_httpd
PING webspace1_httpd (10.0.3.11): 56 data bytes
64 bytes from 10.0.3.11: seq=0 ttl=64 time=0.060 ms
64 bytes from 10.0.3.11: seq=1 ttl=64 time=0.091 ms
64 bytes from 10.0.3.11: seq=2 ttl=64 time=0.083 ms
^C
--- webspace1_httpd ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 0.060/0.078/0.091 ms
```

So caddy plugin uses the wrong ip address?
The backend have two networks attached (proxy, mysql).

[...]

PR #35, opened 2018-05-13
Find container IP on same network as caddy

lucaslorentz:
Fixes https://github.com/lucaslorentz/caddy-docker-proxy/issues/28

Changes in this PR:
- Finds caddy container id, using /proc/self/cgroup
- Get a list of networks attached to caddy container
- When a container IP is needed, it uses IP from the first network in common with caddy container

PS: Nothing changes in swarm service proxy (still uses service dns name).

@pwFoo
Do you think this fixes your issue?


pwFoo:
@lucaslorentz 
> PS: Nothing changes in swarm service proxy (still uses service dns name).
I use docker swarm, but it used IP addresses. Why? Because I use container labels instead of service labels?
So swarm service proxy could fail with dns name if container is attached to multiple networks?

lucaslorentz:
Yes, that's probably because you use container labels.
Swarm service could dns would also fail if not on same network. I will add checks for service network too.

pwFoo:
So with container label it should work. I have to check with a test environment. Thanks!

Issue #53, opened 2018-07-25
multiple caddy environment


hijak:
Hi

I have a situation where we want to use a 2 tier caddy setup in a frontend / backend configuration

Is it possible to introduce some sort of namespaces so we can chuck config to different caddys in the network?

ie

```yaml
---

version: '3.2'
services:
  caddy-front:
   image: jcowey/caddy
   volumes:
     - $(pwd)/caddyfile:/etc/caddy/Caddyfile
    ports:
      - 80:80
      - 443:443
    networks:
      - webnet1

  caddy-backend1:
    image: jcowey/caddy
    networks:
      - webnet2
      - webnet1
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock

  caddy-backend2:
    image: jcowey/caddy
    networks:
      - webnet2
      - webnet1
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock

  app1:
    image: some/image
    networks:
      - webnet2
    labels:
      caddy1.address: apptest1.lvh.me:80
      caddy1.targetport: 80

  app2:
    image: some/image
    networks:
      - webnet2
    labels:
      caddy2.address: apptest2.lvh.me:80
      caddy2.targetport: 80

networks:
  webnet1:
    external: true
  webnet2:
    external: true
```

This will probably envovle starting caddy with a flag to define what id it is
 
hope this makes sense 
 
Regards

hijak:
Or perhaps to configure the loader via env var?

```yaml
caddy-backend1:
    image: jcowey/caddy
    environment:
      - CADDY_DOCKER=caddy1
    networks:
      - webnet2
      - webnet1
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
```

hijak:
awesome thanks

revnelson:
Can you provide an example of how this would work? I can't seem to get it working. I want a main caddy that handles whole-server services like portainer and matomo. Then, in a separate docker-compose, I want a caddy that handles project-specific services like php-fpm, wordpress, etc.

My thought process:

- I can't mount project-specific static files to the main caddy instance (I'd like to be able to easily add/remove/move projects without needing to change the main server configs)
[...]

PR #54, opened 2018-07-25
Adding support for configurable label prefix

jtyr:
This PR is trying to solve the issues #53.

### DIFF
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -9,6 +9,7 @@ import (
 	"html/template"
 	"io/ioutil"
 	"log"
+	"os"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


lucaslorentz:
That's great!

I think environment variables are very appropriate for docker container settings.

But since this is also a caddy plugin, and caddy uses CLI args for options, maybe we should support setting label prefix using a CLI option. Like this:
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/generator.go#L25

What do you think?
Do you think we should support environment variable or cli options or both?

lucaslorentz:
@jtyr 
Can you please add some description to repository README and also #53 ?

jtyr:
Command line option would be very handy as well. I will add it together with the README change.

PR #55, opened 2018-07-25
Adding doc for flags, flag for label prefix and env var for proxy flag

jtyr:
Adding flag for the label prefix and documenting flags as discussed in PR #54.

### DIFF
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -20,24 +20,29 @@ import (
 )
 
 var labelPrefixFlag string
+var labelPrefixEnv string
 var proxyServiceTasksFlag bool
-var isTrue = regexp.MustCompile("^(true|True|TRUE|yes|Yes|YES|1)$")
+var proxyServiceTasksEnv string
+var isTrue = regexp.MustCompile("(?i)^(true|yes|1)$")
 
 func init() {
 	flag.StringVar(&labelPrefixFlag, "docker-label-prefix", "caddy", "Prefix for Docker labels")
 	flag.BoolVar(&proxyServiceTasksFlag, "docker-proxy-service-tasks", false, "Proxy to service tasks instead of VIP")
+
+	labelPrefixEnv = os.Getenv("CADDY_DOCKER_LABEL_PREFIX")
+	proxyServiceTasksEnv = os.Getenv("CADDY_DOCKER_PROXY_SERVICE_TASKS")
 }
 
 func getLabelPrefix() string {
-	if val := os.Getenv("CADDY_DOCKER_LABEL_PREFIX"); val != "" {
-		return val
+	if labelPrefixEnv != "" {
+		return labelPrefixEnv
 	} else {
 		return labelPrefixFlag
 	}
 }
 
 func getProxyServiceTasks() bool {
-	if val := os.Getenv("CADDY_DOCKER_PROXY_SERVICE_TASKS"); isTrue.MatchString(val) {
+	if isTrue.MatchString(proxyServiceTasksEnv) {
 		return true
 	} else {
 		return proxyServiceTasksFlag
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -19,26 +19,36 @@ import (
 	"github.com/docker/docker/client"
 )
 
-var proxyServiceTasks bool
-var caddyNetworks map[string]bool
+var labelPrefixFlag string
+var proxyServiceTasksFlag bool
+var isTrue = regexp.MustCompile("^(true|True|TRUE|yes|Yes|YES|1)$")
 
 func init() {
-	flag.BoolVar(&proxyServiceTasks, "proxy-service-tasks", false, "Proxy to service tasks instead of VIP")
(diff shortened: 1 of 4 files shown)


plugin/generator.go
```diff
@@ -19,26 +19,36 @@ import (
 	"github.com/docker/docker/client"
 )
 
-var proxyServiceTasks bool
-var caddyNetworks map[string]bool
+var labelPrefixFlag string
+var proxyServiceTasksFlag bool
+var isTrue = regexp.MustCompile("^(true|True|TRUE|yes|Yes|YES|1)$")
```
lucaslorentz: Instead of having text with different cases, we could use case insensitive regex:
https://stackoverflow.com/questions/15326421/how-do-i-do-a-case-insensitive-regular-expression-in-go
```
regexp.MustCompile("(?i)^(true|yes|1)$")
```

plugin/generator.go
```diff
@@ -19,26 +19,36 @@ import (
...
 func init() {
-	flag.BoolVar(&proxyServiceTasks, "proxy-service-tasks", false, "Proxy to service tasks instead of VIP")
+	flag.StringVar(&labelPrefixFlag, "docker-label-prefix", "caddy", "Prefix for Docker labels")
+	flag.BoolVar(&proxyServiceTasksFlag, "docker-proxy-service-tasks", false, "Proxy to service tasks instead of VIP")
 }
 
-
-func getCaddyLabelPrefix() string {
+func getLabelPrefix() string {
 	if val := os.Getenv("CADDY_DOCKER_LABEL_PREFIX"); val != "" {
```
lucaslorentz: This could probably be executed only once, inside Init. I don't think the environment variables can change after the process started.

lucaslorentz: This could probably be executed only once, during Init. I don't think the environment variables can change after the process started.

plugin/generator.go
```diff
@@ -19,26 +19,36 @@ import (
...
 		return val
+	} else {
+		return labelPrefixFlag
 	}
+}
 
-	return "caddy"
+func getProxyServiceTasks() bool {
+	if val := os.Getenv("CADDY_DOCKER_PROXY_SERVICE_TASKS"); isTrue.MatchString(val) {
```
lucaslorentz: Same thing here, I think we could compute the value of ProxyServiceTasks only once, during init.

jtyr:
@lucaslorentz please review the last commit.

plugin/generator.go
```diff
@@ -19,26 +19,41 @@ import (
...
+var labelPrefixEnv string
+var proxyServiceTasksFlag bool
+var proxyServiceTasksEnv string
+var isTrue = regexp.MustCompile("(?i)^(true|yes|1)$")
 
 func init() {
-	flag.BoolVar(&proxyServiceTasks, "proxy-service-tasks", false, "Proxy to service tasks instead of VIP")
-}
+	flag.StringVar(&labelPrefixFlag, "docker-label-prefix", "caddy", "Prefix for Docker labels")
+	flag.BoolVar(&proxyServiceTasksFlag, "docker-proxy-service-tasks", false, "Proxy to service tasks instead of VIP")
```
lucaslorentz: This is a breaking change, we should continue using `proxy-service-tasks`

plugin/generator.go
```diff
@@ -19,26 +19,41 @@ import (
...
+	if labelPrefixEnv != "" {
+		return labelPrefixEnv
+	} else {
+		return labelPrefixFlag
 	}
+}
 
-	return "caddy"
+func getProxyServiceTasks() bool {
+	if isTrue.MatchString(proxyServiceTasksEnv) {
```
lucaslorentz: Maybe execute is `isTrue.MatchString` inside init as well?

plugin/generator.go
jtyr: What about to have both for now and remove the old one later? Give people chance to change it? Just mention it in the changelog that it will be removed in next release.

plugin/generator.go
```diff
@@ -19,26 +19,41 @@ import (
...
+		return true
+	} else {
+		return proxyServiceTasksFlag
+	}
 }
 
-var caddyLabelPrefix = getCaddyLabelPrefix()
+var caddyLabelPrefix = getLabelPrefix()
```
lucaslorentz: This is not working, caddyLabelPrefix is empty when generator runs

lucaslorentz:
I had to do some refactorings to generator.go to get cli options working.
I will create a PR into your branch

jtyr:
Thanks for you PR. As an owner of the original repo, you can push your changes directly into my branch I created this PR from. Then your commit will appear in this PR.

lucaslorentz:
Really? I always considered your fork is entirely yours, and that I don't have any write access to it.
I will try to push directly.

jtyr:
Person who if creating PR can choose whether the owner of the original repo can pus to his for (the default). Otherwise when you merging PR, you can choose to squash commits into one which makes your Git history a bit cleaner.

Thanks for polishing the feature and for merging it into your project.

Issue #59, opened 2018-08-16
Constant errors when not using swarm


ptman:
```
dockerd[674]: time="2018-08-16T11:37:52.592586915+02:00" level=error msg="Error getting services: This node is not a swarm manager. Use \"docker swarm init\" or \"docker swarm join\" to connect this node to swarm and try again."
dockerd[674]: time="2018-08-16T11:37:52.593329658+02:00" level=error msg="Handler for GET /v1.35/services returned error: This node is not a swarm manager. Use \"docker swarm init\" or \"docker swarm join\" to connect this node to swarm and try again."
```

Maybe something could be done to limit the requests and therefore error messages? Like only check for swarm once when caddy is starting and then remember it's not available?

lucaslorentz:
Yes, we should do something to limit that.
I think it's possible to check if swarm is enabled via API.
Another option would be to add a flag to disable monitoring swarm services.

ptman:
Excellent, thank you!

PR #60, opened 2018-08-28
Skip services monitoring when swarm is not available

lucaslorentz:
Fixes https://github.com/lucaslorentz/caddy-docker-proxy/issues/59
@ptman what do you think?

I'm not sure if we should check swarm availability only once and cache it.
Or if we should check it every time, that way if the node is disconnected from swarm we nicely handle that with a comment on Caddyfile instead of errors.


lucaslorentz:
I still need to do some local tests.
@ptman
If you would like to test as well, please use the image:
lucaslorentz/caddy-docker-proxy:test

ptman:
If you check availability only once, it could be rechecked with restart or maybe SIGHUP? At least some sort of caching should be done to prevent constant checking, but maybe with a short TTL.

lucaslorentz:
@ptman
I've updated it to cache for 1 minute

PR #61, opened 2018-09-06
Sort directives by their names

jtyr:
This PR implements directives sorting useful when using Caddy [snippets](https://caddyserver.com/docs/caddyfile#snippets) and imports because the snippets must be defined before the directive which is importing it. This is especially useful when the snippets are coming from a different node that the directive. Luckily the snippets always start with `(` which puts them at the top of the sorted list.

### DIFF
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -9,6 +9,7 @@ import (
 	"html/template"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"os"
 	"regexp"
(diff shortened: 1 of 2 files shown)


lucaslorentz:
Thank you.
That makes sense, and it's working.

lucaslorentz:
I'm doing some refactoring on top of it to sort them before writing directives to string.
Along with other improvements to test the full caddyfile generation and mock docker client.

Issue #23, opened 2018-04-06
Add custom config to Caddyfile?


OmgImAlexis:
~Is it possible for example to proxy to a service that isn't within the swarm?~

Renamed the issue as it's more fitting.

lucaslorentz:
Currently we only support generating Caddyfile from service and containers labels.

If that external service is related to one of your swarm services, you can add it's config on that service, like:
```
caddy_0.address=service.example.com
caddy_0.targetport=80
caddy_0.targetpath=/my-path
caddy_1=external.example.com
caddy_1.proxy="/ external-url"
```

If you don't need the proxy to swarm service, you can write only the external one:
```
caddy=external.example.com
caddy.proxy="/ external-url"
```

lucaslorentz:
I think we can easily add support appending the content of a `docker config` to the Caddyfile.

pwFoo:
Your example should also work to create redirects like examples below?
- domain.com -> www.domain.com
- domain1.com -> domain2.com
- domain1.com -> https://www.domain2.com
- ...

PR #64, opened 2018-09-09
Add config content to caddyfile

lucaslorentz:
Fixes: https://github.com/lucaslorentz/caddy-docker-proxy/issues/23

It is available for test on docker image: lucaslorentz/caddy-docker-proxy:test

### DIFF
--- a/examples/service-proxy.yaml
+++ b/examples/service-proxy.yaml
@@ -1,4 +1,10 @@
-version: '3'
+version: '3.7'
+
+configs:
+  caddy-basic-content:
+    file: ./Caddyfile
+    labels:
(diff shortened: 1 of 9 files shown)


examples/service-proxy.yaml
```diff
@@ -1,4 +1,10 @@
-version: '3'
+version: '3.7'
+
+configs:
+  caddy-basic-content:
+    file: ./Caddyfile
+    labels:
+        caddy:
```
lucaslorentz: If the config have the label prefix, we include it's content on the top of the generated caddyfile. Right after warnings comments.

PR #71, opened 2018-09-14
Increase polling interval

lucaslorentz:
Because:
* Docker realtime events are working well to trigger caddy updates.
* Caddyfile generation is bigger now that we check containers, services and configs.

This PR:
* increases the default polling interval from 10s to 30s to save some resources.
* adds CLI flag and environment variable to set polling interval

The polling is a fallback in case an update was not triggered from a docker event.

@jtyr @mrdrup What do you think?

### DIFF
--- a/plugin/loader.go
+++ b/plugin/loader.go
@@ -3,7 +3,9 @@ package plugin
 import (
 	"bytes"
 	"context"
+	"flag"
 	"log"
+	"os"
 	"time"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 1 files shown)


jtyr:
I believe this should be configurable via ENV variable or commandline option.

mrdrup:
Looks good.

Side note. For docker events, it would make sense to aggregate changes during some period of time (should be configurable, with defaults to 3-5s) before triggering caddy reload. This is will prevent constant server reloads when newly deployed containers start crashing.

lucaslorentz:
@jtyr 
I added ENV variable as well: CADDY_DOCKER_POLLING_INTERVAL

@mrdrup
Currently it aggregates it for 100 milliseconds: https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/loader.go#L105

On my tests I found that same action might trigger multiple events, and aggregating for 100 milliseconds avoid double updates for the same action.

I personally think 3s to 5s it's a bit too much for real time reactions.  I'm willing to increase the default to 500ms and also make it configurable.

PR #79, opened 2018-10-23
Add caddyfile path for default config

### LINKED ISSUE #78: Custom config without docker swarm
What do you thing about adding a way to merge a default Caddyfile that we can mount in container so we can use this without swarm ?

This will allow to add default config like redirect http to https.

Its-Alex:
Add cli command and env variable to use a default CaddyFile

Fix #78

### DIFF
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"log"
 	"os"
 	"regexp"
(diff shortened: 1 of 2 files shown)


Its-Alex:
@lucaslorentz I can't make it works with go 1.11.1, can you provide some advices to test it ? Thanks

Its-Alex:
Up ?

lucaslorentz:
LGTM

plugin/generator.go
```diff
@@ -105,6 +117,19 @@ func (g *CaddyfileGenerator) GenerateCaddyFile() []byte {
 
 	directives := map[string]*directiveData{}
 
+	if g.caddyFilePath != "" {
+		dat, err := ioutil.ReadFile(g.caddyFilePath)
+		if err != nil {
+			if _, err := buffer.Write(dat) {
```
crvv: This line can't compile.
Do you mean `if _, err := buffer.Write(dat); err {}` ?

Its-Alex: Yes, thanks! I made a PR to solve it #82

PR #82, opened 2018-11-13
Fix typo in generator

Its-Alex:
In my previous pull request there is a mistake that I fix in this one


lucaslorentz:
Thanks for fixing it.
I've just enabled pull requests builds as well.

lucaslorentz:
Btw, it was still broken. :-)
Fixing it now.

PR #84, opened 2018-11-13
Fix build

lucaslorentz:
@Its-Alex 
Can you please review?

We should create unit tests later for that, mocking file system access.


Its-Alex:
Looks good to me

Its-Alex:
You must declare error outside of if statement to get it works

lucaslorentz:
Well, I will fix it later :-)
Working without go compiler here.

Issue #85, opened 2018-11-13
add option to disable telemetry


sheychen290:

lucaslorentz:
We don't affect caddy cli flags.
I think you can just follow caddy documentation on how to disable it.
https://caddyserver.com/docs/telemetry

lucaslorentz:
https://caddyserver.com/docs/cli#disabled-metrics

thoom:
The only way to completely disable telemetry is at compile time by setting the flag `enableTelemetry` to false. Would it be possible to generate a docker image with this disabled?

lucaslorentz:
Yes, I will work on it

thoom:
I was thinking about adding a PR for this. What would you expect the resulting Docker images to be named? i.e. `caddy-docker-proxy:alpine-no-telemetry`? Or would you rather move the default to no telemetry and instead `caddy-docker-proxy:alpine` is with telemetry disabled and `caddy-docker-proxy:alpine-telemetry` is what is currently built?

thoom:
Or maybe just disable it for `caddy-docker-proxy` based on this comment: https://github.com/mholt/caddy/issues/2355#issuecomment-440072106?

lucaslorentz:
I think we should go a bit more towards adding a cli option + env variable to disable telemetry instead of doubling the number of tags we have in docker image.

lucaslorentz:
I'm still not sure if it should be enabled or disabled by default.
@mholt Maybe you can provide some feedback about the feature, will it be discontinued?

mholt:
We don't have any plans to discontinue telemetry. We might change its defaults in Caddy 2, but Caddy 2 will have a monitoring platform as well so we will have to see how that changes things!

EarthlingDavey:
I know this issue was closed a while ago, but it's still coming up as a to google result. 

As a workaround for docker, the telemetry url can be effectively blocked with an extra host argument of `--add-host telemetry.caddyserver.com:127.0.0.1` or for docker-compose:

```
extra_hosts:
 - "telemetry.caddyserver.com:127.0.0.1"
```

francislavoie:
Caddy v2 has no telemetry.

PR #89, opened 2018-12-13
Added builds that have telemetry disabled

thoom:
Add no-telemetry builds for all of the various `caddy-docker-proxy` images. Fixes https://github.com/lucaslorentz/caddy-docker-proxy/issues/85.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,14 @@ RUN apk add -U --no-cache ca-certificates
 FROM scratch
 LABEL maintainer "Lucas Lorentz <lucaslorentzlara@hotmail.com>"
 
+ARG CADDY_EXEC=caddy 
+
 EXPOSE 80 443 2015
 ENV HOME /root
 
 WORKDIR /
 COPY --from=alpine /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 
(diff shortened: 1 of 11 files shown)


thoom:
I'm posting this as a POC and feedback on implementation, didn't think about it running the automated build.

thoom:
Looks like this breaks because of this is tied to caddy `0.11.0` and we can't disable telemetry in the source code until `0.11.1`, released a few weeks ago.

lucaslorentz:
Hey @thoom 
Sorry for the very late feedback.

I've updated to caddy 1.0.

About this PR. I'm still a bit concerned about doubling docker tags because of **no-telemetry**.

Maybe we can create a new cli flag like: `disable-telemetry`. And use it's value to set `caddymain.EnableTelemetry`.  We usually add env variables as well, like: https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/generator.go#L74

lucaslorentz:
@thoom 
What do you think?
Would you like to update the PR?

lucaslorentz:
Fixed on https://github.com/lucaslorentz/caddy-docker-proxy/pull/100

thoom:
Sorry for the late reply, I think PR #100 is the right solution. I don’t remember my original reason for creating different docker images, but it was definitively not elegant ☺️

PR #92, opened 2019-01-25
Fake NULL Values

mcrosson:
Some container management solutions don't allow NULL or empty string label values. An example is portainer (https://github.com/portainer/portainer/issues/2646).

This patch creates a faux NULL value that's replaced during label processing to ensure that NULL value parameters can be passed through the parsing engine despite upstream not allowing such.

This has been tested with Portainer 1.2.0 using a custom caddy build to ensure correctness.

The value chosen is ```[0xNULLx0]``` and that's wholly invalid generally speaking per the Caddyfile syntax.

### DIFF
--- a/plugin/generator.go
+++ b/plugin/generator.go
@@ -416,13 +416,13 @@ func processVariables(data interface{}, content string) string {
 }
 
 func parseArgs(text string) []string {
+	text = strings.Replace(text, "[0xNULLx0]", "", -1)
 	args := regSplit(text, "\\s+")
 	if len(args) == 1 && args[0] == "" {
 		return []string{}
 	}
 	return args
 }
-
 func regSplit(text string, delimeter string) []string {
 	reg := regexp.MustCompile(delimeter)
 	indexes := reg.FindAllStringIndex(text, -1)


lucaslorentz:
@mcrosson  That's nice!
I think it would be nice to cover it with at least 1 test though.
I was also wondering, we already support go templates inside label values.
So, maybe instead of having a new feature, we could come up with a template interpolation that results in empty values.

lucaslorentz:
@mcrosson 
Ha! Guess what. I tested and those 2 options already works:
`{{\"\"}}`
`{{nil}}`
I prefer the second one.
Can you please try to use that with Portainer?
If it works well, we can document it in the README

mcrosson:
I will give it a try when I'm back near my test machine.

Is there any documentation on using go templates inside the labels?

lucaslorentz:
Actually it is not documented.

There are tests ensuring it works well:
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/generator_test.go#L341

mcrosson:
@lucaslorentz Looks like ```{{nil}}``` works with portainer.

I just deployed FileBrowser container and the ```transparent``` is lacking any value yet the container label is ```"caddy.proxy.transparent": "{{nil}}",```

PR #122, opened 2019-08-24
Validate each service independently

### LINKED ISSUE #120: [Feature request] validate each service independently.
### Problem

If a service is configured incorrectly, then caddy will keep running with the last working configuration (via USR1) and log the error (good).

Mispelling "websocket" for example

```
labels:
  caddy.targetport: 8080
  caddy.address: example.com
  caddy.proxy.websockett: ""
```

`CaddyFile error: :40 - Error during parsing: unknown property 'websockett'`

But if that error goes unnoticed, and the caddy container restarts, caddy won't run because it starts with a clean state and then can't fallback with USR1.

### Possible solution

For each service, create a temporary caddyfile (in memory I imagine), run caddy with the validate flag, log the errors if there is any, and then only stitch together the services that works. Now a misconfigured service won't take all the other services down.

Because you can apply a global caddyfile with caddy-docker-proxy, that needs to be inserted as well for each service when it validates.

I had a discussion about this over on the caddy forum, and gamalan came up with this idea which I think would be really great to implement.
https://caddy.community/t/fault-tolerance-docker-clustering/6097

lucaslorentz:
Fix #120

Working on validating each service independently.
I still need to run some tests, but it seems quite feasible.

@Maxiride

### DIFF
--- a/plugin/caddyfileprocessor.go
+++ b/plugin/caddyfileprocessor.go
@@ -1,0 +1,109 @@
+package plugin
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"sort"
+	"strings"
+
+	"github.com/caddyserver/caddy"
+	"github.com/caddyserver/caddy/caddyfile"
+
+	_ "github.com/caddyserver/caddy/caddyhttp" // plug in the HTTP server type
+)
+
+// ProcessCaddyfile validate and removes wrong server blocks from caddyfile
+func ProcessCaddyfile(caddyfileContent []byte) []byte {
+	serverBlocks, err := caddyfile.Parse("", bytes.NewReader(caddyfileContent), nil)
+
+	if err != nil {
+		log.Printf("[ERROR] Error parsing caddyfile:%s\n", err)
+	}
+
+	var validServerBlocks []caddyfile.ServerBlock
+	for _, serverBlock := range serverBlocks {
+		serverBlockContent := serializeServerBlock(&serverBlock)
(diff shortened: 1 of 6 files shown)


plugin/caddyfileprocessor.go
```diff
@@ -0,0 +1,105 @@
...
+
+	for _, directiveName := range getSortedDirectiveNames(serverBlock.Tokens) {
+		indent := 1
+		newLine := true
+		tokenLine := -1
+		for _, token := range serverBlock.Tokens[directiveName] {
+			if token.Text == "}" {
+				indent--
+			}
+			if token.Line != tokenLine {
```
lucaslorentz: @mholt
Strange that caddyfile.Parse doesn't return token for line breaks inside directives. So I need to rely on the original token line to recognize line breaks.

```
proxy / service2:5000/api {
  except /a /b
  transparent
}
```
Return tokens:
```
["proxy", "/", "service2:5000/api", "{", "/a", "/b", "transparent", "}"]
```
Or something like that

mholt: Yeah, the v2 parser (already committed to Caddy's v2 branch) cleans this up a bit by introducing something called "segments" which were always really there implicitly. A segment starts with a directive and goes until the end of the line or the end of the block that line opens. That's what is happening here.

You will find working with tokens is much easier if you use a Dispenser (in the caddyfile package). Same goes for v2 when that lands.

PR #137, opened 2020-05-17
Caddy v2

### LINKED ISSUE #130: Caddy v2?
Is this project going to remain targeting v1 alone, or is there plans to support v2 (currently in Beta, aimed to release early 2020) at a later date?

lucaslorentz:
Fixes #130 

Huge PR. It has a lot of refactorings as well. 

I think changes to README file is a good starting point, as it describes the behavior changes.


README.md
```diff
@@ -360,24 +356,24 @@ docker stack rm caddy-docker-demo
...
-curl -H Host:whoami1.example.com http://localhost:2015
+curl -k --resolve whoami0.example.com:443:127.0.0.1 https://whoami0.example.com
+curl -k --resolve whoami1.example.com:443:127.0.0.1 https://whoami1.example.com
 
 docker rm -f caddy whoami0 whoami1
 ```
 
 ## Building it
-You can use our caddy build wrapper **build.sh** and include additional plugins on https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/main.go#L5
+You can use our caddy build wrapper **build.sh** and include additional plugins on https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/main.go#L8
```
francislavoie, line 372: I think you could point people to use https://github.com/caddyserver/xcaddy instead. There's some docs on the site for it:

https://caddyserver.com/docs/build#xcaddy

Could do the building that way in your Docker image possibly as well (multi-stage, first stage is Go plus xcaddy, then 2nd stage you just copy the binary)

azure-pipelines.yml
```diff
@@ -5,10 +5,10 @@ jobs:
 
   variables:
     GOBIN:  '$(GOPATH)/bin' # Go binaries path
-    GOROOT: '/usr/local/go1.13' # Go installation path
+    GOROOT: '/usr/local/go1.14' # Go installation path
     GOPATH: '$(system.defaultWorkingDirectory)/gopath' # Go workspace path
     modulePath: '$(GOPATH)/src/github.com/$(build.repository.name)' # Path to the module's code
-    # TODO: modules should be the default in Go 1.13, so this won't be needed
+    # TODO: modules should be the default in Go 1.14, so this won't be needed
```
francislavoie, line 11: That's right, I don't think you need to specify it here.

P.S. If you want some help to set up CI with GitHub Actions, let me know 👍 

See https://github.com/caddyserver/caddy/blob/master/.github/workflows/ci.yml

README.md
```diff
@@ -185,31 +203,10 @@ Proxying multiple domains to container
...
-admin.example.com {
-	proxy / servicename:81
-}
-```
-
 ### Docker configs
 You can also add raw text to your caddyfile using docker configs. Just add caddy label prefix to your configs and the whole config content will be prepended to the generated caddyfile.
 
-[Here is an example](examples/service-proxy.yaml#L4)
+[Here is an example](examples/example.yaml#L4)
```
francislavoie, line 210: (This comment isn't really meant for this line)

I didn't see anything in the README about global options and how they're handled.

I think a global options label would be something like `caddy_0={{""}}` right?

What happens if two containers try to set the same options? I'm thinking maybe global options should only be allowed to be set on the caddy-docker-proxy container (maybe as labels, whatever you think is best) to avoid any potential conflicts.

plugin/caddyfile/fromlabels_test.go
```diff
@@ -0,0 +1,114 @@
...
+	const expectedCaddyfile = "localhost {\n" +
+		"	group {\n" +
+		"		a value-a\n" +
+		"		b value-b\n" +
+		"		group {\n" +
+		"			a value-a\n" +
+		"			b value-b\n" +
+		"		}\n" +
+		"	}\n" +
+		"}\n"
```
francislavoie, line 29: Couldn't you just write this with ` template strings? A lot less Go syntax soup that way (applies for all the expected strings)

plugin/caddyfile/processor.go
```diff
@@ -1,43 +1,39 @@
...
-	serverBlocks, err := caddyfile.Parse("", bytes.NewReader(caddyfileContent), nil)
+// Process caddyfile and removes wrong server blocks
+func Process(caddyfileContent []byte) ([]byte, []byte) {
+	logsBuffer := bytes.Buffer{}
+
+	serverBlocks, err := caddyfile.Parse("", caddyfileContent)
 
 	if err != nil {
-		log.Printf("[ERROR] Error parsing caddyfile:%s\n", err)
+		logsBuffer.WriteString(fmt.Sprintf("[ERROR] Error parsing caddyfile: %s\n", err.Error()))
```
francislavoie, line 19: In v2 you can use `zap` for logging. Here's a note from the docs: https://caddyserver.com/docs/extending-caddy#logs

plugin/cmd.go
```diff
@@ -0,0 +1,97 @@
...
+
+func init() {
+	caddycmd.RegisterCommand(caddycmd.Command{
+		Name:  "docker-proxy",
+		Func:  cmdFunc,
+		Usage: "<command>",
+		Short: "Run caddy as a docker proxy",
+		Flags: func() *flag.FlagSet {
+			fs := flag.NewFlagSet("docker-proxy", flag.ExitOnError)
+			fs.String("caddyfile-path", "", "Path to a base CaddyFile that will be extended with docker sites")
```
francislavoie, line 26:
```suggestion
			fs.String("caddyfile-path", "", "Path to a base Caddyfile that will be extended with docker sites")
```

plugin/cmd.go
```diff
@@ -0,0 +1,97 @@
...
+	validateNetworkFlag := flags.Bool("validate-network")
+	processCaddyfileFlag := flags.Bool("process-caddyfile")
+	pollingIntervalFlag := flags.Duration("polling-interval")
+
+	options := &config.Options{}
+
+	if caddyFilePathEnv := os.Getenv("CADDY_DOCKER_CADDYFILE_PATH"); caddyFilePathEnv != "" {
+		options.CaddyFilePath = caddyFilePathEnv
+	} else {
+		options.CaddyFilePath = caddyFilePath
```
francislavoie, line 58: In general, I think a quick cleanup of `CaddyFile` -> `Caddyfile` would be good 😄

README.md
```diff
@@ -269,43 +266,41 @@ Be aware that this needs to be tested further.
...
+  -caddyfile-path string
+    	Path to a base CaddyFile that will be extended with docker sites
+  -label-prefix string
+    	Prefix for Docker labels (default "caddy")
+  -polling-interval duration
+    	Interval caddy should manually check docker for a new caddyfile (default 30s)
+  -process-caddyfile
+    	Process Caddyfile before loading it, removing invalid servers
+  -proxy-service-tasks
+    	Proxy to service tasks instead of service load balancer
```
francislavoie, line 287: This is probably just because of my inexperience with swarm, but what's the difference?

plugin/generator/generator_test.go
```diff
@@ -0,0 +1,144 @@
...
+	dockerClient.ServicesData = []swarm.Service{
+		{
+			Spec: swarm.ServiceSpec{
+				Annotations: swarm.Annotations{
+					Name: "service",
+					Labels: map[string]string{
+						"caddy_version":  "0.11.0",
+						"caddyversion":   "0.11.0",
+						"caddy_.version": "0.11.0",
+						"version_caddy":  "0.11.0",
```
francislavoie, line 68: Should maybe update these to `2.0.0`?

Dockerfile-alpine
```diff
@@ -1,10 +1,12 @@
 FROM alpine:3.10
 LABEL maintainer "Lucas Lorentz <lucaslorentzlara@hotmail.com>"
```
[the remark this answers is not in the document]
francislavoie: lol what the heck, nevermind... I typed this review comment on my phone when waking up, I think it might be a bug in the github mobile app that was hiding `< >` strings 😱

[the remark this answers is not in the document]
francislavoie: Proof that I'm not insane 😂: 

image

README.md
```diff
@@ -137,23 +152,26 @@ You can customize the automatic generated proxy with the following special label
 When all the values above are added to a service, the following configuration will be generated:
 ```
 service.example.com {
-	proxy /source https://servicename:8080/api
+  route /source/* {
+    uri strip_prefix /source
+    rewrite * /api{uri}
```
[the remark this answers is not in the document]
lucaslorentz: > Hmm, I don't think this is exactly correct - rewrites apply to the path, not the uri.

I changed to /api{path} and it works well, code pushed.

> Also this specific example could be done with a uri replace instead of strip_prefix then rewrite. I think if both source and target are set, you can take that approach instead.

Wouldn't `uri replace` replace the text in any part of the uri/path?
I need to make sure it replaces on in the beginning.

> Also /source/* won't match requests to /source btw, so maybe you'd want to add a redir rule outside the route to do redir /source /source/ 308 ?

Good point, I was trying to to keep the same behavior as caddy v1 `proxy /source upstream`.
But I'm not sure if caddy v1 would also match /source or not. Do you know that by heart?

When necessary, users can do a rewrite by themselves like this example: https://github.com/lucaslorentz/caddy-docker-proxy/blob/caddy-v2/examples/example.yaml#L88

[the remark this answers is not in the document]
francislavoie: Wait, I think I'm wrong, I think `{uri}` is right 🤦‍♂️ sorry (but I think both work ultimately)

https://github.com/caddyserver/caddy/blob/master/modules/caddyhttp/rewrite/rewrite.go#L207

plugin/caddyfile/processor.go
lucaslorentz: I played a bit with zap on another branch. But I have a specific requirement here and I didn't find how to do with it.

Because we do polling and monitor all events from docker, we generate caddyfile several times but if it is the same as the last generated one, we discard the logs and new caddyfile.

Thus, the logsBuffer is there. To store logs without actually printing them.

Do you have any suggestions?

I think this can be done in another PR later.

francislavoie: Hmm, maybe make like a TempLog struct, something like `Level, Msg` and append to it, then later emit all the logs with zap at once?

You could override the log level for each one instead of discarding, setting them all to `debug`. That way, if a user sets the `debug` global option, they'll get a bit more info that might help with figuring out why nothing happened when they spun up a new container 😄

README.md
lucaslorentz: When you create a service in swarm, it creates a Virtual IP with a load balancer that balances to all replicas. Tasks are containers executing in order to fulfill a specific service replica configuration.

By default, service VIP will be registered as upstream in Caddy, thus Caddy will be doing load balancing, only routing.

If you chose to proxy to tasks, all running instances of a service are registered in Caddy, and VIP is skipped, thus Caddy load balancer and health checks are used.

Proxying to VIP was more stable and was the first implemented option. Maybe we should change the default to proxy to tasks now.

francislavoie: Interesting! Thanks for the explanation.

BTW, Caddy v2 supports SRV DNS as well, I don't think it matters in this case but thought I'd mention it, different way to do the same thing.

README.md
```diff
@@ -317,11 +317,11 @@ You can modify docker connection using the following environment variables:
 ## Volumes
 On a production docker swarm cluster, it's **very important** to store Caddy folder on a persistent storage. Otherwise Caddy will re-issue certificates every time it is restarted, exceeding let's encrypt quota.
 
-To do that map a docker volume to `/root/.caddy` folder.
+To do that map a persistent docker volume to `/data` folder.
```
francislavoie, line 320:
```suggestion
To do that, map a persistent docker volume to `/data` folder.
```

README.md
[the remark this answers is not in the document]
lucaslorentz: > Wait, I think I'm wrong, I think {uri} is right 🤦‍♂️ sorry (but I think both work ultimately)

{path} looks good. Indeed it supports both, as I don't want to change query, {path} is more appropriate.

README.md
```diff
@@ -338,18 +333,19 @@ Clone this repository.
```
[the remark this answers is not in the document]
lucaslorentz: Good point, changed CADDY default paths on all dockerfiles and readme.

README.md
```diff
@@ -269,43 +266,41 @@ Be aware that this needs to be tested further.
 
 This is an example of how to mount the windows docker pipe using CLI:
 ```
-docker run --rm -it -p 2015:2015 -v //./pipe/docker_engine://./pipe/docker_engine lucaslorentz/caddy-docker-proxy:ci-nanoserver-1803 -agree -email email@example.com -log stdout
+docker run --rm -it -p 2019:2019 -v //./pipe/docker_engine://./pipe/docker_engine lucaslorentz/caddy-docker-proxy:ci-nanoserver-1803 -agree -email email@example.com -log stdout
```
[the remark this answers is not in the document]
lucaslorentz: Removed

README.md
lucaslorentz: I noticed that. Unfortunately, Docker doesn't create SRV records for internal DNS. AFAIK.

README.md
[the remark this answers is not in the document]
mholt: > Wouldn't uri replace replace the text in any part of the uri/path?
> I need to make sure it replaces on in the beginning.

Yep, replace operates on both the path and the raw query string.

README.md
```diff
@@ -269,43 +266,41 @@ Be aware that this needs to be tested further.
...
 ## Caddy CLI
-All flags and environment variables supported by Caddy CLI are also supported:
-https://caddyserver.com/docs/cli
-
-Check **examples** folder to see how to set them on a docker compose file.
-
-This plugin provides these flags:
+This plugin extends caddy cli with command `caddy docker-proxy` and flags.
 
+Run `caddy docker-proxy --help` to see all available flags:
```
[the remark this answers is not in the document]
mholt: I also think this would be useful -- anticipating lots of questions from users. But doesn't have to block this PR.

README.md
mholt: Seconding this; xcaddy is probably the best way to make custom Caddy builds.

banjerluke:
This is great! I'm just setting up my first Docker Swarm cluster and have been tearing my hair out trying to get HA Traefik running. Discovered Caddy a couple days ago and loved what I saw, but was disappointed to find out that v2 didn't have Docker integration yet. Then stumbled on this today. Thank you for doing this! I'm eager to try it out when it's ready. 👏👏👏 😁

plugin/caddyfile/processor.go
mholt: I wonder if a better way to do this is to implement a log writer that can buffer/discard logs as it sees fit?

mholt:
This is very impressive. I did a quick first pass and didn't notice anything obviously wrong. As far as the overall approach, I'd say if it works for your requirements, then what you have here looks good to me.

plugin/caddyfile/fromlabels_test.go
lucaslorentz: Yes. I tried several times with ``. Basically, I would like to use `` but keep indentation instead of moving everything to left.
Tried with this: https://github.com/lithammer/dedent
But it was always adding one extra line at top.

francislavoie: You could also put the test data in files in a `testdata` dir, would be easier to maintain I think. I could help with doing that if you like

plugin/caddyfile/processor.go
lucaslorentz: Agree, I will merge without that first.

lucaslorentz: https://github.com/lucaslorentz/caddy-docker-proxy/issues/139

README.md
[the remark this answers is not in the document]
lucaslorentz: https://github.com/lucaslorentz/caddy-docker-proxy/issues/140

README.md
lucaslorentz: https://github.com/lucaslorentz/caddy-docker-proxy/issues/141

README.md
lucaslorentz: Everything in the README up to **Caddy label args creates a server block** are global options.
Global diretive:
```
caddy.tls = internal
```

Site-block directive:
```
caddy = localhost
caddy.tls = internal
```

lucaslorentz: I will try to describe it better

francislavoie: Hmm, that seems a bit confusing, I think maybe something like `caddy_global.admin = off` might feel better, instead of relying on the order in which labels are defined 🤔

plugin/caddyfile/fromlabels_test.go
lucaslorentz: Yes, sure, help is very welcome!
I think it's good somehow to keep test input and expected output close to each other.
Maybe moving input labels also into files would solve that...
Or implementing some deindentation of ` ` strings.

francislavoie: For Caddy I took this sort of approach:

https://github.com/caddyserver/caddy/blob/master/caddytest/integration/caddyfile_adapt/log_roll_days.txt

Single files with a bunch of dashes as a split delimiter, first part is Caddyfile, 2nd part is JSON.

Could do the same but with labels first, Caddyfile 2nd.

azure-pipelines.yml
lucaslorentz: Cleaned up that old stuff.
Interesting, I think we can migrate to Github Actions
We need to think as well as how to reduce building steps in the project.
Using xcaddy would improve it a bit, but it would be nice if we can remove all Dockerfiles from this project as well.

PR #142, opened 2020-05-19
Refactor tests to use test files for input/expected pairs

francislavoie:
As discussed previously, this refactors the tests that use labels or simple strings as inputs into files that contains the input/expected pairs.

I didn't bother touching the tests that have structs as input, I don't have a good way to clean those up while keeping the input/expected together.

This test strategy is something I set up for https://github.com/caddyserver/caddy, it's pretty much the same code for parsing the files. I also pulled the envfile parsing function that's in there as well, with a few modifications like trimming both sides and allowing whitespace around the `=`, because labels aren't as strict as env vars.

### DIFF
--- a/plugin/caddyfile/fromlabels_test.go
+++ b/plugin/caddyfile/fromlabels_test.go
@@ -1,175 +1,106 @@
 package caddyfile
 
 import (
+	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
 	"testing"
 	"text/template"
 
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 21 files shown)


lucaslorentz:
I think we can do the same for tests heavily based on docker data:
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/generator/containers_test.go#L12

Probably we can unmarshal a json into `dockerClient`.
Maybe for a separate PR for that.

Btw, next build will pass:
https://github.com/lucaslorentz/caddy-docker-proxy/commit/1a998db62b245293131052129eae5a78c9cbb519

Issue #20, opened 2018-03-16
Security problem with mounted docker.sock?


pwFoo:
Because docker.sock ist mountet to the caddy reverse proxy it could be a security issue? What do you think about? Would it possible to split it up in a caddy reverse proxy and a second container with a shared caddyfile / caddy config?

lucaslorentz:
I'm not a security expert. But this is what I think:

I don't think it is a security issue. Of course you should avoid mounting docker.sock to other containers. Only mount it on containers you trust and know it needs connection to docker. Like this project, that completely relies on docker host.

Currently we have 2 docker image variants, from alpine and from scratch. The scratch image not even have a shell installed, I recommend using that one if you're concerned about security.

I also want to mention that although mounting docker.sock is the most convenient way to run it, there are alternatives. If your container is capable of accessing the docker host by IP, you could use the connection environment variables (DOCKER_HOST...) to setup a connection.

And last, can you please explain a bit further your suggestion? Do you wan't to separate the caddyfile generation logic, from the caddy web server? So, one container would generate the caddyfile, and multiple others would use it to serve http trafic using caddy?

pwFoo:
Hi @lucaslorentz and thanks for quick reply. Separate cadddyfile generation logic (with access to docker.sock) from the reverse proxy / caddy (serve http traffic) should be more secure I think. But from scratch without installed shell / linux root filesystem should be fine! So I just have to change the tag from alpine to scratch? I'll try it.

lucaslorentz:
Actually, just remove alpine. Scratch is the default one.
And I recommend not using the CI version, use a released one.
Like 0.1 to allow patch updates.
Or 0.1.1 to lock to a specific version.

lucaslorentz:
@pwFoo 
Scratch image seems to be a secure option for you.
Please, reopen this issue if you think otherwise.

deeky666:
I still have headaches with the docker socket mounted into the Caddy container. Also this means containers can only run on manager nodes (swarm), right? This is somewhat a limitation if you need to use replication mode `host` (in case you need to inspect the real client ip from requests).
Wouldn't it be possible to separate the plugin into a dedicated service and then have a shared Caddy folder that is used by the plugin and the server? The plugin service could write the generated Caddyfile to the shared folder then to make it available for all Caddy server instances.
[...]

PR #144, opened 2020-05-21
Add option to run controller separate from server

lucaslorentz:
Fixes https://github.com/lucaslorentz/caddy-docker-proxy/issues/20


tests/distributed/compose.yaml
```diff
@@ -0,0 +1,89 @@
+version: '3.7'
+
+services:
+
+  caddy_server:
+    image: caddy-docker-proxy:local
```
lucaslorentz, line 6: @francislavoie @mholt
Is there any simple way to run easily run official docker image with Admin API exposed to *:2019?
Maybe something like `caddy admin-api --bind ":2019"`.

If it is possible, I will consider reducing this project to be only a caddy server controller, that discovers all caddy servers inside docker cluster and configures them.

francislavoie: Hmm, that's a good question. I don't think we have anything in place for that now other than having an initial config as a base. Like make sure the initial Caddyfile has the admin global option set to `0.0.0.0:2019`.

I think we could add an env variable that would set the admin address (if it's not already set in a config file, in which case the config file would win).

mholt: First, are you sure that's a good idea? That means any client on any interface can read and change the config.

But, to answer your question, you can change the admin endpoint by setting the `listen` property of the `admin` struct in your config: https://caddyserver.com/docs/json/admin/#listen

There's no command line flag for this since all config is in the config file. The only thing command line flags are for is bootstrapping the config or customizing the command itself. So you can use the `--config` flag to pass a config, even if it only sets that one field, then it will use that value at startup as if it was specified on the command line directly.

francislavoie: @mholt it's fine to do it in this case because of Docker networking. Only containers in the same network could access it.

plugin/generator/generator.go
```diff
@@ -155,7 +179,15 @@ func (g *CaddyfileGenerator) GenerateCaddyfile() ([]byte, string) {
 		logsBuffer.Write(processLogs)
 	}
 
-	return caddyfileContent, logsBuffer.String()
+	if len(caddyfileContent) == 0 {
+		caddyfileContent = []byte("# Empty caddyfile")
```
francislavoie, line 183: Why this instead of empty string?

mholt: I think if Caddy reads an empty file it returns an EOF error?

francislavoie: Makes sense.

PR #150, opened 2020-05-25
Isolate directives when different order is defined

lucaslorentz:
At first, I wasn't isolating directives with different order mainly because there were directives generated automatically.

Now that we use {{upstreams}} and all directives are explicitly defined by the user. We can safely change the behavior so that a different order also isolates directives.

That new behavior will be more intuitive in cases like:
https://github.com/lucaslorentz/caddy-docker-proxy/issues/149


lucaslorentz:
The advantage as well is that you would never need to define prefix and suffix together.
You just need to choose:
a. If order matters for me, then I will add a prefix.
b. I want to isolate, but don't care about the order, then I will add a suffix.

francislavoie:
Very nice! Thanks for this one 😁 should definitely reduce confusion

Issue #141, opened 2020-05-18
Switch build to use xcaddy


lucaslorentz:
I think you could point people to use https://github.com/caddyserver/xcaddy instead. There's some docs on the site for it:

https://caddyserver.com/docs/build#xcaddy

Could do the building that way in your Docker image possibly as well (multi-stage, first stage is Go plus xcaddy, then 2nd stage you just copy the binary)

_Originally posted by @francislavoie in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3NDEzMTYwMDk4/pull_request_reviews/more_threads_

mholt:
Feel free to ping me if you run into any issues with xcaddy.

lucaslorentz:
```
xcaddy build --with github.com/lucaslorentz/caddy-docker-proxy/plugin/v2
```
Works well!

While changing build to use xcaddy we can probably go back to a single package, removing main.
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/go.mod
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/go.mod

lucaslorentz:
@mholt @francislavoie  
I just started working on this and faced a minor issue.
I would like to use XCaddy to cross-compile for ARM while being on AMD64. Or compile for windows while being on Linux.
There is a check after compilation that I need to disable because the output is not executable in the build architecture/OS.
https://github.com/caddyserver/xcaddy/blob/master/cmd/xcaddy/main.go#L131
I will open an issue on XCaddy for now.

mholt:
@lucaslorentz Should be fixed now.

PR #151, opened 2020-05-27
Switch build to x-caddy

lucaslorentz:


go.mod
```diff
@@ -1,10 +0,0 @@
-module github.com/lucaslorentz/caddy-docker-proxy/v2
```
francislavoie, line 1: Don't you still need the `go.mod` so others can install this as a plugin?

francislavoie: Oh nevermind, scratch that. I see https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/go.mod :sweat_smile:

lucaslorentz: There is a go.mod inside plugin folder.

lucaslorentz:
Fixed https://github.com/lucaslorentz/caddy-docker-proxy/issues/141

PR #160, opened 2020-07-04
Fix snippets being lost when configured in configs or caddyfile

### LINKED ISSUE #159: Snippets created in Caddyfile aren't accessible in Docker Labels
Whenever you create a snippet in a caddyfile and try to import it with a docker label, the application segfaults and exits.
I'm not sure if this is intended or unintended behavior. 

## Reproduce:
- Caddyfile has a snippet, and container is set with CADDY_DOCKER_CADDYFILE_PATH.
- You use the snippet in another docker-compose label `caddy.import: "<snippet>"`.

On startup, the program throws a invalid memory address error and a segfault. 

## Workaround:
- Caddyfile doesn't have snippet, instead you write the snippet in a docker-compose caddy label.
- You use the Snippet in another docker-compose label.

Snippet gets properly imported on startup.

## Stacktrace:
```caddy    | panic: runtime error: invalid memory address or nil pointer dereference
caddy    | [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x13289f6]
caddy    | 
caddy    | goroutine 62 [running]:
caddy    | github.com/lucaslorentz/caddy-docker-proxy/plugin/v2/caddyfile.Process(0xc000423880, 0x342, 0x342, 0x657, 0x342, 0x0, 0x0, 0x0, 0x0)
caddy    |      github.com/lucaslorentz/caddy-docker-proxy/plugin/v2@v2.0.0-20200616192750-9ce6e7859b70/caddyfile/processor.go:21 +0xa6
caddy    | github.com/lucaslorentz/caddy-docker-proxy/plugin/v2/generator.(*CaddyfileGenerator).GenerateCaddyfile(0xc000181e00, 0x6fc23ac00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
caddy    |      github.com/lucaslorentz/caddy-docker-proxy/plugin/v2@v2.0.0-20200616192750-9ce6e7859b70/generator/generator.go:192 +0xa40
caddy    | github.com/lucaslorentz/caddy-docker-proxy/plugin/v2.(*DockerLoader).update(0xc0002d7680, 0x0)
caddy    |      github.com/lucaslorentz/caddy-docker-proxy/plugin/v2@v2.0.0-20200616192750-9ce6e7859b70/loader.go:134 +0x6c
caddy    | github.com/lucaslorentz/caddy-docker-proxy/plugin/v2.(*DockerLoader).Start.func1()
caddy    |      github.com/lucaslorentz/caddy-docker-proxy/plugin/v2@v2.0.0-20200616192750-9ce6e7859b70/loader.go:81 +0x2a
caddy    | created by time.goFunc
caddy    |      time/sleep.go:168 +0x44
```

lucaslorentz:


plugin/caddyfile/lexer.go
```diff
@@ -0,0 +1,170 @@
...
+	"io"
+	"unicode"
+)
+
+type (
+	// lexer is a utility which can get values, token by
+	// token, from a Reader. A token is a word, and tokens
+	// are separated by whitespace. A word can be enclosed
+	// in quotes if it contains whitespace.
+	lexer struct {
```
lucaslorentz, line 28: @francislavoie @mholt
Had to copy lexer from caddy repo.
I'm fine with it, and lexer features don't change that often.

Does it make sense to expose in caddy a Parse function that preserves environment variables and snippets? Or maybe expose lexer publicly?

lucaslorentz: Just to give more context:
- I parse several caddyfiles, from different sources.
- Then I merge them into a single caddyfile, putting global options first, then snippets then sites.

Because snippet declaration and import might be in different input caddyfiles, I must be able to do the parsing step without processing snippet and environment variables.

francislavoie: I think this should do what you need? https://github.com/caddyserver/caddy/pull/3549

lucaslorentz: Thanks. That's pretty much what I need.

I don't know the implications though, for example, by making it public probably caddy shouldn't make breaking changes on it. So, I'm completely fine if it's best for Caddy to keep the token concept an internal thing.

I will wait next caddy release before cleaning up duplicated code.

francislavoie: Yeah for sure. I need Matt's +1 on it first.

mholt: Dang, copying the whole lexer is going to be a pain to maintain. I've tried to avoid exporting it because then the burden is on us to make sure it doesn't change and then we have to maintain the API, and most people want the parser instead of the lexer anyway.

Your case is a bit niche... hmm. 

We might be able to get away with exporting it if we put a big scary warning in the godoc saying that this API is not stable.

lucaslorentz: Indeed. Another option is to add a "parsePreservingSnippets" function. That parses it to server blocks, and each snippet would also be a "serverBlock".

PR #166, opened 2020-07-17
clarifications to example yaml files

happysalada:
I have noticed, my automated yaml formatted made some changes.
I can revert those if not needed.

I have also grouped two changes into one commit.
- There seems to be a typo in the distributed example file
- I added some comments mostly to the standalone example file.

### DIFF
--- a/examples/efs-volume.yaml
+++ b/examples/efs-volume.yaml
@@ -1,4 +1,4 @@
-version: '3'
+version: "3"
 
 services:
   caddy:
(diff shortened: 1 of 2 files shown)


francislavoie:
FYI Caddy will automatically fallback to the staging endpoint if the first attempt on the live LE endpoint fails, and switches back to the live one once using the staging endpoint succeeds. This help avoid rate limit issues. See https://caddyserver.com/docs/automatic-https#errors

That said, it's still a good idea to use the staging endpoint like you noted there. Just wanted to say that it's not black-and-white for that part.

PR #170, opened 2020-07-22
Use tokenizer to parse label quotes into tokens

### LINKED ISSUE #168: multiline directives seems not to work
Using caddy2 by utilizing docker 
```caddy:
    image: lucaslorentz/caddy-docker-proxy:latest
    container_name: caddy
    restart: unless-stopped
    command: docker-proxy  -caddyfile-path=/config/Caddyfile
    networks:
      proxy:
    ports:
      - "80:80"
      - "443:443"
    environment:
      - MY_DOMAIN=$MYDOMAIN
    volumes:
      - caddy_data/:/data
      - caddy_config/:/config
      - /var/run/docker.sock:/var/run/docker.sock
    labels:
      - caddy.email=$MYMAIL
```
I tried to get most caddy configuration in the individual docker-compose files. For nextcloud I should add some header information. I tried (note the two differnt way to add escape a quote by using escape and double quotes and single escape quotes :
```
    labels:
      caddy: nc.MYDOMAIN nextcloud.MYDOMAIN
      caddy.header: "/"
      caddy.header.Strict-Transport-Security: "\"max-age=15552000;\""
      caddy.header.Referrer-Policy: "strict-origin-when-cross-origin"
      caddy.header.X-XSS-Protection: \"1; mode=block\"
      caddy.header.X-Content-Type-Options: "nosniff"
      caddy.header.X-Frame-Options: "SAMEORIGIN"
      caddy.reverse_proxy: "{{upstreams 80}}"
```
Actually I tried many forms with and without quotes, with escaped quotes, etc. But nothing seems to work correctly.
I always checked in the caddy logs, how it was parsed.
E.g. the above example was interpreted as
```
nc.MYDOMAIN nextcloud.MYDOMAIN {
       header / {
                 Referrer-Policy strict-origin-when-cross-origin
                 Strict-Transport-Security "\"max-age=15552000;\""
                 X-Content-Type-Options nosniff
                 X-Frame-Options SAMEORIGIN
                 X-XSS-Protection "\"1;" "mode=block\""
             }
}
```
As you can see, there are those funny double quotes now, were I expect that one quote should be gone, like in the single quote cases.

However, not escaping those quotes, they are removed completely:
```
header / {
     Referrer-Policy strict-origin-when-cross-origin
     Strict-Transport-Security max-age=15552000;
     X-Content-Type-Options nosniff
     X-Frame-Options SAMEORIGIN
     X-XSS-Protection 1; mode=block
}
```
Thus, I am a bit puzzled how to get a single set of quotes

lucaslorentz:
Improvements for #168


plugin/caddyfile/testdata/labels/quotes.txt
```diff
@@ -0,0 +1,12 @@
...
+caddy.0_with_spaces    = "a b c d"
+caddy.1_without_spaces = "abcd"
+caddy.2_multiple       = "a b c d" "abcd"
+caddy.3_back           = `a b c d`
+----------
+localhost {
+	with_spaces "a b c d"
+	without_spaces abcd
+	multiple "a b c d" abcd
+	back "a b c d"
```
francislavoie, line 11: This isn't right I don't think. It should preserve the backticks. They're usually used intentionally to wrap JSON text like:

```
caddy.3_back = `{"some":"json"}`
```

lucaslorentz: I updated that test.
During parsing, it parses into a single token. I don't think we need to preserve the original quote used to define that token.

But when serializing that token back, we need to make sure it generates a valid Caddyfile, and today it does:
https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/plugin/caddyfile/marshal.go#L44

lucaslorentz: If the token has whitespace, or quotes, it is printed back to caddyfile inside double quotes and any quotes are escaped.

francislavoie: :+1:

francislavoie:
How will it work for yaml if you need multiple args with or without quotes? Should backticks be used in yaml in general do you think?

plugin/caddyfile/marshal.go
```diff
@@ -41,7 +41,13 @@ func (block *Block) write(buffer *bytes.Buffer, level int) {
 			buffer.WriteString(" ")
 		}
 
-		if strings.ContainsAny(key, ` "'`) {
+		if strings.ContainsAny(key, "\n\"") {
```
lucaslorentz, line 44: @francislavoie 
Updated this logic, to make use of backticks whenever it can to improve readability

plugin/caddyfile/testdata/labels/quotes.txt
```diff
@@ -0,0 +1,18 @@
...
+caddy.4_escaped        = "a\"b"
+caddy.5_unbalanced     = "a
+caddy.6_unbalanced     = a"
+----------
+localhost {
+	with_spaces "a b c d"
+	without_spaces abcd
+	multiple "a b c d" abcd
+	back `{"some":"json"}`
+	escaped `a"b`
```
lucaslorentz, line 15: @francislavoie 
This is now a case, where we would not preserve the original quotes used to declare the token.
But the generator knows backticks generate a more readable string because it doesn't require scaping.

lucaslorentz: @francislavoie 
What do you think?

francislavoie: I guess I'm indifferent? Feels like an edgecase that won't come up so often. :man_shrugging:

lucaslorentz:
I think you can chose between double-quotes or backticks.
All those options should work now:
```yaml
mylabel: "\"arg 1\" \"arg 2\""
mylabel: "`arg 1` `arg 2`"
mylabel: '"arg 1" "arg 2"'
mylabel: '`arg\n1` "arg 2"'
```

I'm not sure about these ones:
```yaml
mylabel: |
  "arg 1" "arg 2"
mylabel: |
  `line1
  line2` otherarg
```

francislavoie:
Ah right, single quotes helps the situation :+1:

lucaslorentz:
Wow, according to https://yaml-online-parser.appspot.com/ that works very well. I'm surprised!

lucaslorentz:
@francislavoie 
In caddyfile, can backtick be used to declare multiline strings? Like:
```
respond / `
<html>
....
</html>
`
```

francislavoie:
Yeah, multiline backticks works:

```json
{
	"apps": {
		"http": {
			"servers": {
				"srv0": {
					"listen": [
						":80"
					],
					"routes": [
						{
							"match": [
								{
									"path": [
										"/"
									]
								}
							],
							"handle": [
								{
									"body": "\n\u003chtml\u003e\n....\n\u003c/html\u003e\n",
									"handler": "static_response"
								}
							]
						}
					]
				}
			}
		}
	}
}
```

Issue #175, opened 2020-08-10
What do you think about adding a Prometheus exporter plugin to this image?


flaviostutz:
I could work on that.

lucaslorentz:
I'm not sure.
Each user has different needs regarding plugins, I don't see how to decide which plugins we should have by default.
Maybe we should focus on simplifying how to build a custom docker image with the plugins you need.

If there is a simple way to build your own image with the plugins you want, do you think that would be enough?

lucaslorentz:
@flaviostutz 
Could you please take a look on https://github.com/lucaslorentz/caddy-docker-proxy/pull/176 ?

flaviostutz:
The design of #176 is great. I will discuss more there. Thanks for the fast reply!

PR #176, opened 2020-08-10
Add Dockerfile to be called from docker build

lucaslorentz:
How to use it before merging this PR:
```
docker build --build-arg "PLUGINS=github.com/caddyserver/nginx-adapter github.com/hairyhenderson/caddyprom" -t your-tag-name https://github.com/lucaslorentz/caddy-docker-proxy.git#builder
```

After merging this PR, we can drop the `#builder` (branch).

@flaviostutz
What do you think about this as as fix for https://github.com/lucaslorentz/caddy-docker-proxy/issues/175?

PS: I still need to udpate the README file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,13 @@
-FROM alpine:3.11 as alpine
-RUN apk add -U --no-cache ca-certificates
+ARG CADDY_VERSION=2.1.1
 
-# Image starts here
-FROM scratch
-LABEL maintainer "Lucas Lorentz <lucaslorentzlara@hotmail.com>"
+FROM caddy:${CADDY_VERSION}-builder AS builder
 
-EXPOSE 80 443 2019
-ENV XDG_CONFIG_HOME /config
-ENV XDG_DATA_HOME /data
+ARG PLUGINS=
 
-WORKDIR /
-COPY --from=alpine /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+RUN caddy-builder \
+    github.com/lucaslorentz/caddy-docker-proxy/plugin/v2 \
+    ${PLUGINS}
 
-COPY artifacts/binaries/linux/amd64/caddy /bin/
+FROM caddy:${CADDY_VERSION}
 
-ENTRYPOINT ["/bin/caddy"]
-
-CMD ["docker-proxy"]
+COPY --from=builder /usr/bin/caddy /usr/bin/caddy
--- a/Dockerfile-scratch
+++ b/Dockerfile-scratch
@@ -1,0 +1,19 @@
+FROM alpine:3.11 as alpine
+RUN apk add -U --no-cache ca-certificates
+
+# Image starts here
+FROM scratch
+LABEL maintainer "Lucas Lorentz <lucaslorentzlara@hotmail.com>"
+
+EXPOSE 80 443 2019
+ENV XDG_CONFIG_HOME /config
+ENV XDG_DATA_HOME /data
+
+WORKDIR /
+COPY --from=alpine /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+
+COPY artifacts/binaries/linux/amd64/caddy /bin/
+
+ENTRYPOINT ["/bin/caddy"]
+
+CMD ["docker-proxy"]
--- a/build-images-linux.sh
+++ b/build-images-linux.sh
@@ -3,7 +3,7 @@
 set -e
 
 chmod +x artifacts/binaries/linux/amd64/caddy
-docker build -t lucaslorentz/caddy-docker-proxy:ci -f Dockerfile .
+docker build -t lucaslorentz/caddy-docker-proxy:ci -f Dockerfile-scratch .
 docker build -t lucaslorentz/caddy-docker-proxy:ci-alpine -f Dockerfile-alpine .
 
 chmod +x artifacts/binaries/linux/arm32v6/caddy


francislavoie:
FWIW we probably plan to move the official docker image to use `xcaddy` instead eventually. We might leave the `caddy-builder` script in for backwards compatibility though. /cc @hairyhenderson

lucaslorentz:
@francislavoie
That will be a nice improvement.

@hairyhenderson
Do you think this could be done in the official caddy docker repo?

Create a Dockerfile in the root of https://github.com/caddyserver/caddy-docker, which accepts the plugins and caddy version as args. [Like this](https://github.com/lucaslorentz/caddy-docker-proxy/blob/builder/Dockerfile)

That way it would be even simpler to build a custom image, just one command:
```
docker build --build-arg "PLUGINS=plugin1 plugin2" https://github.com/caddyserver/caddy-docker
```

flaviostutz:
I think this is a good resolution for #175.

hairyhenderson:
On the topic of moving to `xcaddy` in the official image, I just logged https://github.com/caddyserver/caddy-docker/issues/107

As for a `docker build` that uses args, that reminds me of what Grafana supports: https://grafana.com/docs/grafana/latest/installation/docker/#build-and-run-a-docker-image-with-pre-installed-plugins - they also have the option of [installing them at runtime](https://grafana.com/docs/grafana/latest/installation/docker/#install-plugins-in-the-docker-container) which I've used, but is a little awkward.

Personally I prefer just using a straight `Dockerfile`, because then there's no ambiguity - the build is self-contained and reproducible, whereas with build-args you would need to save the command somewhere to make sure you have the right mix of plugins. And if you're going to save a shell script with a `docker build` command, why not just save a `Dockerfile`?

lucaslorentz:
@hairyhenderson
Exactly! I saw it the first time in grafana, I couldn't remember that anymore.
There is nothing more appealing than short and simple instructions in docs, that's probably why they adopted that approach.

Dockerfile is simple, but it is undeniably one more step for users, while they could just point to a remote Dockerfile.

I will add the Dockerfile to this repo.

If Caddy officially adopts this way as well in the future I will be happy to point users to official Caddy instructions.

hairyhenderson:
> Dockerfile is simple, but it is undeniably one more step for users, while they could just point to a remote Dockerfile.

IMO it's the same number of steps: _either_ users must commit a build script to their source control, _or_ they must commit a `Dockerfile`. Though I suppose the former only requires running a shell script, whereas the latter does require knowing to run a `docker build` command. 🤔

Personally I've been teaching people to write Dockerfiles instead of shell scripts due to the improvements in portability and reproducibility, but there's nothing wrong with providing alternate methods.

I'll consider the `build-args` approach for the `caddy-docker` repo, but it is going to be a bit more complex than simply pointing at `https://github.com/caddyserver/caddy-docker`, because we need to be able to support multiple different Dockerfiles for different versions of Caddy. Plus it also means that users of this method would _not_ be using the official `caddy:builder` image, but instead rebuilding it themselves.

lucaslorentz:
@hairyhenderson
Have you seen the Dockerfile I created in this PR?
https://github.com/lucaslorentz/caddy-docker-proxy/blob/b6d2b7e3486d96252d61f41fc69e985ba426e950/Dockerfile

I did versioning as just another argument and it still uses caddy:builder.
I basically copied the example from official caddy docker docs and added 2 args + a default plugin.
It is working fine by just running:
```
docker build --build-arg "PLUGINS=github.com/hairyhenderson/caddyprom" https://github.com/lucaslorentz/caddy-docker-proxy.git#builder
```

lucaslorentz:
I would suggest maintaining only one Dockerfile like that.
And add more arguments for architecture/os, that way people can use windows images as well, and so on.

hairyhenderson:
Hmm! Thanks @lucaslorentz - I hadn't actually read the Dockerfile that closely (still on vacation so not paying enough attention 😅). It is actually pretty straightforward.

I think the only blocker for supporting it in the caddy-docker repo is that repo is focused on the [official image](https://hub.docker.com/_/caddy), and official images have some restrictions, one of which is that build-args are not allowed (due to requirements on reproducibility). I've opened https://github.com/caddyserver/caddy-docker/issues/108 to continue that discussion, though.

francislavoie:
I think this can be closed? The official builder image now uses xcaddy so I don't think this is still necessary?

Dockerfile
```diff
@@ -1,19 +1,13 @@
...
-EXPOSE 80 443 2019
-ENV XDG_CONFIG_HOME /config
-ENV XDG_DATA_HOME /data
+ARG PLUGINS=
 
-WORKDIR /
-COPY --from=alpine /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+RUN caddy-builder \
+    github.com/lucaslorentz/caddy-docker-proxy/plugin/v2 \
+    ${PLUGINS}
```
francislavoie, line 9:
```suggestion
RUN xcaddy build \
    --with github.com/lucaslorentz/caddy-docker-proxy/plugin/v2 \
    ${PLUGINS}
```

francislavoie: The `${PLUGINS}` part might need to be revised to run a shell command to prefix each one with `--with` to be compatible with `xcaddy`.

Dockerfile
```diff
@@ -1,19 +1,13 @@
-FROM alpine:3.11 as alpine
-RUN apk add -U --no-cache ca-certificates
+ARG CADDY_VERSION=2.1.1
```
francislavoie, line 1:
```suggestion
ARG CADDY_VERSION=2.4.0
```

Dockerfile
```diff
@@ -1,19 +1,13 @@
...
+    ${PLUGINS}
 
-COPY artifacts/binaries/linux/amd64/caddy /bin/
+FROM caddy:${CADDY_VERSION}
 
-ENTRYPOINT ["/bin/caddy"]
-
-CMD ["docker-proxy"]
\ No newline at end of file
+COPY --from=builder /usr/bin/caddy /usr/bin/caddy
```
francislavoie, line 13: It'll still be necessary to override the `CMD`

```suggestion
COPY --from=builder /usr/bin/caddy /usr/bin/caddy

CMD ["caddy", "docker-proxy"]
```

lucaslorentz:
Thanks @francislavoie 
I will abandon this PR, my recommendation is to follow Caddy Docs and use official caddy images

Issue #218, opened 2021-02-14
Docker image tag and arch mismatch


0uep:
There is an inconsistency between the image tag and its OS/arch.

TAG | OS/ARCH | Docker image details
---|---|--
ci-arm32v6 | linux/amd64 | [hub.docker.com/layers/lucaslorentz/caddy-docker-proxy/ci-arm32v6](https://hub.docker.com/layers/lucaslorentz/caddy-docker-proxy/ci-arm32v6/images/sha256-49ce809fceb78a6d9a3dc2106d95d03a90361de449c9b853faf63a1ae7008696)
2.3-arm32v6 | linux/amd64 | [hub.docker.com/layers/lucaslorentz/caddy-docker-proxy/2.3-arm32v6](https://hub.docker.com/layers/lucaslorentz/caddy-docker-proxy/2.3-arm32v6/images/sha256-49ce809fceb78a6d9a3dc2106d95d03a90361de449c9b853faf63a1ae7008696)
2.3.4-arm32v6 | linux/amd64 | [hub.docker.com/layers/lucaslorentz/caddy-docker-proxy/2.3.4-arm32v6](https://hub.docker.com/layers/lucaslorentz/caddy-docker-proxy/2.3.4-arm32v6/images/sha256-49ce809fceb78a6d9a3dc2106d95d03a90361de449c9b853faf63a1ae7008696)

lucaslorentz:
Hey @0uep 
Thanks for reporting this.
Can you please double check if I put the right platform ids in https://github.com/lucaslorentz/caddy-docker-proxy/pull/219 ?

lucaslorentz:
I'm facing issues with the build pipeline while fixing that tag mismatch.

As a workaround, please use the alpine variants until this issue is solved:
- ci-alpine-arm32v6
- 2.3-alpine-arm32v6
- 2.3.4-alpine-arm32v6

0uep:
Hey @lucaslorentz
Your change seems OK, but not sure. Hope it will fix the issue. :crossed_fingers:
Thank you for your super work.
I have integrated it in my build (see my Dockerfile https://github.com/lucaslorentz/caddy-docker-proxy/pull/219#issuecomment-786070532).
However I have not yet migrated my Caddyfile to my docker-compose.yml…

0uep:
I have wrote my previous message at the same time as yours.
Hope you will fix it easily. :)

PR #219, opened 2021-02-15
Switch to docker multiplatform images and update packages

lucaslorentz:
Fix https://github.com/lucaslorentz/caddy-docker-proxy/issues/218


francislavoie:
Maybe you could consider switching to the `caddy:<version>-builder` image variant now? https://hub.docker.com/_/caddy

0uep:
Hi @francislavoie 

I was not aware about `caddy:<version>-builder`. For your information, this is the way I build my docker image including three plugins:

```Dockerfile
FROM golang

RUN set -eux \
    ; \
    mkdir -p \
        /config/caddy \
        /data/caddy \
        /etc/caddy \
        /etc/gatekeeper/auth/local \
        /usr/share/caddy \
    ; \
    GO111MODULE=on go get -u github.com/caddyserver/xcaddy/cmd/xcaddy \
    ; \
    xcaddy build --with github.com/greenpau/caddy-auth-portal \
                 --with github.com/greenpau/caddy-auth-jwt \
                 --output /usr/bin/caddy

# set up nsswitch.conf for Go's "netgo" implementation
# - https://github.com/docker-library/golang/blob/1eb096131592bcbc90aa3b97471811c798a93573/1.14/alpine3.12/Dockerfile#L9
RUN [ -e /etc/nsswitch.conf ] || echo 'hosts: files dns' > /etc/nsswitch.conf

# https://caddyserver.com/docs/conventions#file-locations
ENV XDG_CONFIG_HOME /config
ENV XDG_DATA_HOME   /data

VOLUME /config
VOLUME /data
VOLUME /etc/caddy
VOLUME /etc/gatekeeper/auth

# Let's Encrypt TLS challenge
EXPOSE 80

# Official static web site
EXPOSE 443

# /metrics endpoint for Prometheus
EXPOSE 2019

WORKDIR /srv

ENV TZ "UTC0"
CMD ["caddy", "run", "--config", "/etc/caddy/Caddyfile", "--adapter", "caddyfile"]
```

lucaslorentz:
CDP is executed using a different CLI command.

Your last line should be something like:
```
CMD ["caddy", "docker-proxy"]
```

0uep:
Thank you very much @lucaslorentz 
I will check that when I will be motivated to migrate Caddyfile -> docker label

lucaslorentz:
@francislavoie Tried to use caddy builder image, but it would increase too much the build time because each platform would need to download all golang modules.
I will stick with building all binaries directly in the build agent for now.

PR #237, opened 2021-04-22
Send load commands to previous server, not new

regbo:


lucaslorentz:
Closing this for now.
It would be nice if we were able to reproduce the issue in a consistent way to investigate further. It's not clear to me what is actually hapenning.

PR #296, opened 2021-09-19
Added support for h2c protocol to upstreams

durandj:
Currently the only protocols that are supported by this plugin are
`http` and `https` but Caddy itself does have support for `h2c` as well.

### DIFF
--- a/plugin/generator/labels.go
+++ b/plugin/generator/labels.go
@@ -33,6 +33,9 @@ func labelsToCaddyfile(labels map[string]string, templateData interface{}, getTa
 		"https": func() string {
 			return "https"
 		},
+		"h2c": func() string {
+			return "h2c"
+		},
 	}
 
 	return caddyfile.FromLabels(labels, templateData, funcMap)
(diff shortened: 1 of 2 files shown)


francislavoie:
For ref, the `h2c` scheme shortcut was added in this PR: https://github.com/caddyserver/caddy/pull/3629

durandj:
Wow thanks for the super fast review!

Looks like the build failed though for a reason completely unrelated. Do you have any idea what's going on with that error?

francislavoie:
It's not your fault -- the built tooling needs to be updated to use Go 1.17, on this line:

https://github.com/lucaslorentz/caddy-docker-proxy/blob/4893c4dbcfbdfefca0a6aa6510d79a42536e6c8a/azure-pipelines.yml#L9

durandj:
Seems like the location that Azure Pipelines is pulling the Go installation from doesn't have anything past 1.16.3. Working to figure out what's going on with that.

durandj:
Reverted and rebased and all things are green!

Thanks for the workaround and for getting things working!

You both are amazing :)

PR #303, opened 2021-10-14
Multiple docker sockets support

amaelFr:
I tried to add an implementation for multiple docker sockets, related to the problem I opened on the topic.

I used your implementation and only changed the docker client to an array (and associated variables like jump events).

Could you please take a look at it, if it meets your principles?

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,6 +482,13 @@ Usage of docker-proxy:
   -ingress-networks string
         Comma separated name of ingress networks connecting caddy servers to containers.
         When not defined, networks attached to controller container are considered ingress networks
+  -docker-sockets
+        Comma separated docker sockets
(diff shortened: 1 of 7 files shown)


amaelFr:
After several searches, it appears to be an error in my code, but it is mitigated by calling the update function further down in the code.

The two instance of `append' could be replace by a line of code like
```go
dockerLoader.skipEvents = make([]bool, len(dockerLoader.dockerClients))

```

Do you want me to update the code ?

lucaslorentz:
> Do you want me to update the code ?

Yes, sure. The one line instead of 2 appends seems better.

amaelFr:
I have updated the part of the code concerned by your review.

PR #354, opened 2022-04-19
Update windows images and try to run integration tests on windows

### LINKED ISSUE #333: Nanoserver images not working, missing netapi32.dll
It appears that, sometime between the last time I used this project for Windows containers and now (I'd assume most likely during the upgrade to Caddy v2), there is now a dependency on netapi32.dll which does not exist in Nanoserver container images.  The end result is that the images no longer with Windows containers.

One solution to this would be to switch from using Nanoserver to using Server Core, though this has the issue of adding approximately 2 GB to the container image size.  Alternatively, another solution I've seen from Stefan Scherer is to pull both the Nanoserver and Server Core images, then copy the netapi32.dll file from the Server Core image, as shown in https://github.com/StefanScherer/dockerfiles-windows/blob/main/traefik/Dockerfile.1809#L3.

lucaslorentz:
Update windows images, fix #333 and try to get integration tests running on windows containers


TheJayMann:
I've noticed that these checks take a long time to build, and I would guess this is due to the long time it takes to download and extract the servercore docker image.  It perhaps could build faster if the base nanoserver matched the build server (matching nanoserver:1809 with windows-2019, and nanoserver:ltsc2022 with windows-2022), and then copying the netapi32 from the host itself instead of taking it from the corresponding servercore docker image.

lucaslorentz:
After MANY attempts I got integration tests running for windows, and I was able to reproduce the netapi32 error in the build server:
https://dev.azure.com/lucaslorentzlara/lucaslorentzlara/_build/results?buildId=841&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=3cfa5d13-d456-5c32-1f87-e23b82c00683&l=48

lucaslorentz:
@TheJayMann I will try to copy it from the build server.
I'm a bit outdated regarding windows base images.
Which variants would you recommend to support? 1809 and ltsc2022?
Do you think it's necessary to have both nanoserver and servercore images?

TheJayMann:
Given that Windows Server 2016 has so many limitations regarding containers, and that it is fairly unlikely that somebody will be running a semi-annual Windows Server release, I would say having the nanoserver 1809 and ltsc2022 (which is process isolation compatible with Windows Server 2019 and Windows Server 2022 long term release, respectively) is good enough, assuming the nanoserver image has netapi32 copied over.  In my testing using nanoserver with netapi32 copied, there is no need to have a server core image.

Given that caddy docker proxy in so far as I've seen it used typically only runs caddy and the docker proxy plugin with no other software, there is no real need to have a server core image.  Were it the case that it was common to copy other applications in to the image and run them, and those applications often required server core image because it is unable to run in nanoserver, then I could see a justification for also including server core images.

PR #367, opened 2022-05-18
Update build.sh

francislavoie:
I was just poking around and I noticed this inconsistency. Was this a mistake, or on purpose? All the other `xcaddy build` commands didn't use `/v2`.


lucaslorentz:
I think v2 was required at some moment. Before I fixed the module name.

PR #368, opened 2022-05-19
Add root module and remove plugin module

### LINKED ISSUE #366: Xcaddy not pulling 2.6.0
Currently not able to build a new caddy image and pull down v2.6.0.  Image alwasy ends up with 2.0.0.

`docker exec -it caddy /bin/sh -c 'caddy build-info | grep proxy'`
```
github.com/lucaslorentz/caddy-docker-proxy/plugin/v2 v2.0.0-20211030110928-6ec85e1d3732 h1:w28Q1e39fnrDBi+68t5WOUaX/bUL9Kv6kBa4l0xKY0U=
```

**Dockerfile**
```
ARG CADDY_VERSION=2.5.1

FROM caddy:${CADDY_VERSION}-builder AS builder
#FROM caddy:builder AS builder

RUN xcaddy build \
    --with github.com/greenpau/caddy-security \
    --with github.com/greenpau/caddy-trace \
    --with github.com/porech/caddy-maxmind-geolocation \
    --with github.com/caddy-dns/cloudflare \
    --with github.com/kirsch33/realip

FROM caddy:${CADDY_VERSION}-alpine
#FROM caddy:builder-alpine

RUN apk add --no-cache tzdata

CMD ["caddy", "docker-proxy"]
```

lucaslorentz:
Fix #366 

I had a look at https://github.com/golang/go/wiki/Modules and found that we're doing the following things wrong:

1. It is not advised to have submodules like we have inside /plugin folder.
2. Version tags for submodules must have the module name as prefix. `github.com/lucaslorentz/caddy-docker-proxy/plugin@v2.4.0` would only work if we create tags as `plugin/v2.4.0`, instead of how we create today `v2.4.0`
3. Modules should have /v2 suffix in their name, and also all imports should have /v2 in the package name

This diff creates a new module and removes the old submodule.

We will need to instruct users to stop using `github.com/lucaslorentz/caddy-docker-proxy/plugin`, and use `github.com/lucaslorentz/caddy-docker-proxy/v2` or `github.com/lucaslorentz/caddy-docker-proxy/v2@2.7.0`.

I still need to investigate if the old import `github.com/lucaslorentz/caddy-docker-proxy/plugin` will continue working and will get the 2.6.0 version. That way we don't break anyone's builds.


lucaslorentz:
Moved source files out of plugin folder.
If the old import `github.com/lucaslorentz/caddy-docker-proxy/plugin` stop working. I can just add a module inside plugin folder that imports `github.com/lucaslorentz/caddy-docker-proxy/v2`, just for backwards compatibility.

francislavoie:
We should probably put a big notice at the top of the README mentioning the change in module name

francislavoie (approved):
Makes sense to me 👍

lucaslorentz:
@francislavoie 
Or I could just keep everything as it is and create release tags like `plugin/v2.6.0`. I would only need to adjust the build to recognize those tags.
Maybe this is a better approach. What do you think?

francislavoie:
🤷‍♂️ I'm not sure what's best, I think whatever's easiest for you to manage.

lucaslorentz:
Ok, I will keep the submodule for now.
I will fix how we release the submodule.
Later we can still move to a root module if we want.

PR #403, opened 2022-09-25
Add dependabot

lucaslorentz:
Adding dependabot to test it, maybe it will make it even simpler to keep caddy up to date


lucaslorentz:
Funny that the build failed, looks like master build is broken. #402 will probably fix it.

PR #405, opened 2022-09-25
Update packages further by reinitializing go.mod

lucaslorentz:


lucaslorentz:
Looks like this is not required, and I did on top of the old commit as well

PR #407, opened 2022-09-30
Update README about build image

bytemain:
Change the plugin name and sync content from Dockerfile-alpine

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,18 +567,27 @@ If you need additional Caddy plugins, or need to use a specific version of Caddy
 The main difference from the instructions on the official image is that you must override `CMD` to have the container run using the `caddy docker-proxy` command provided by this plugin.
 
 ```Dockerfile
-ARG CADDY_VERSION=2.5.0
+ARG CADDY_VERSION=2.6.1
 FROM caddy:${CADDY_VERSION}-builder AS builder
 
 RUN xcaddy build \
-    --with github.com/lucaslorentz/caddy-docker-proxy/plugin \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -567,18 +567,27 @@ If you need additional Caddy plugins, or need to use a specific version of Caddy
...
     --with <additional-plugins>
 
 FROM caddy:${CADDY_VERSION}-alpine
 
+EXPOSE 80 443 2019
+
+ENV XDG_CONFIG_HOME /config
+ENV XDG_DATA_HOME /data
+
+RUN apk add -U --no-cache ca-certificates curl
```
francislavoie, line 584: These are all unnecessary/redundant, the base image already sets these appropriately.

README.md
```diff
@@ -567,18 +567,27 @@ If you need additional Caddy plugins, or need to use a specific version of Caddy
...
+
 COPY --from=builder /usr/bin/caddy /usr/bin/caddy
 
-CMD ["caddy", "docker-proxy"]
+ENTRYPOINT ["/bin/caddy"]
+
+CMD ["docker-proxy"]
```
francislavoie, line 590: I don't agree with this change, this makes it harder to get a shell into the container with `/bin/sh` to muck around and debug when necessary.

bytemain: Thanks for your advice, I agree with that.

by the way, these configs is copied from the `Dockerfile-alpine`: https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/Dockerfile-alpine#L13

lucaslorentz:
Thank you @bytemain

PR #408, opened 2022-09-30
Update Readme#Caddy CLI to use double-hyphen

Majr25:
Caddy 2.6.0 does not support single-hyphen long options https://github.com/caddyserver/caddy/pull/4565


francislavoie (approved):
Thanks!

PR #410, opened 2022-10-07
Add ability to not send scopes as a filter with event listener

JakeMakesStuff:
Unfortunately, doing so in Podman will sometimes lead to an EOF, causing the proxy to stop consuming events. This solves that issue with Podman compatibility.


JakeMakesStuff:
Reopening from my workplaces org

PR #415, opened 2022-10-14
Add ability to not send scopes as a filter with event listener

JakeMakesStuff:
Unfortunately, doing so in Podman will sometimes lead to an EOF, causing the proxy to stop consuming events. This solves that issue with Podman compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -521,6 +521,7 @@ CADDY_DOCKER_MODE=<string>
 CADDY_DOCKER_POLLING_INTERVAL=<duration>
 CADDY_DOCKER_PROCESS_CADDYFILE=<bool>
 CADDY_DOCKER_PROXY_SERVICE_TASKS=<bool>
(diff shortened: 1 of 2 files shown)


lucaslorentz:
Nice! Fix looks good!

But I think it will reach more users if we either:
- Detect podman automatically using dockerClient, maybe doing some call to get server info.
- Or rename the env variable to "CADDY_DOCKER_PODMAN" or "CADDY_DOCKER_PODMAN_COMPAT".

What do you think?

lucaslorentz:
I think `dockerClient.Info(context.Background())` should give you the server name.
We've done something similar in https://github.com/lucaslorentz/caddy-docker-proxy/blob/27652e80a618ce26e5bdd8be4235082e8a1f1dd9/generator/generator.go#L217

Another option is detecting swarm instead of podman. If swarm is not available there is no point in using scopes filter.

PR #431, opened 2022-11-28
Define controller networks using the network name

### LINKED ISSUE #286: Can I use the network name instead of IP range to define controller network?
Rather than reserve a specific network range i think it would be much more straightforward to be able to configure the caddy controller something like this:

```
      - CADDY_CONTROLLER_NETWORK_NAME=caddy-controller
...
networks:
    caddy-controller:
        internal: true
```
Is this possible? It should then be able to figure out the range automatically.

lucaslorentz:
Currently,  there is only 1 way to configure CDP in distributed mode (controllers + servers):
- Setting `CADDY_CONTROLLER_NETWORK` with network CIDR on both controller and servers 

What I'm changing:
- `CADDY_CONTROLLER_NETWORK` can be configured with the network name when configured in controllers, or completely omitted if the controller is connected to a single network.
- `CADDY_CONTROLLER_NETWORK` can be replaced by `CADDY_CONTROLLER_URL` on servers, and servers will fetch the controller network information from a new endpoint exposed in controllers.
- When both `CADDY_CONTROLLER_NETWORK` and `CADDY_CONTROLLER_URL` are omitted on servers, they will allow configuration from all networks they're connected to.

This simplifies configuration for distributed mode and removes the need to hardcode network CIDRs:
- When having a single network for ingress and controlling, you don't have to configure any of those settings anymore. Keep in mind that this approach is less secure, as any proxied container will have access to CDP servers' admin endpoints.
- When separating ingress and controlling networks, the easiest configuration now is setting CADDY_CONTROLLER_NETWORK with the network name in controllers and setting CADDY_CONTROLLER_URL on servers.

Fixes #286


generator/networkselector.go
```diff
@@ -0,0 +1,58 @@
+package generator
+
+import "net"
+
+type NetworkSelector struct {
```
lucaslorentz, line 5: Thinking about renaming this, maybe NetworkInfo

generator/networkselector.go
```diff
@@ -0,0 +1,58 @@
...
+	ID      string
+	Name    string
+	Subnets []net.IPNet
+}
+
+type NetworkSelectors struct {
```
lucaslorentz, line 11: Thinking about renaming this, maybe NetworkGroup

PR #449, opened 2023-01-11
Adding HTTP Basic Authentication Documentation

nahakiole:
Adding an example for HTTP Basic Authentication. I've looked for this when I started using the project and couldn't find it. Now with the added documentation I finally understood the format 😅, so I thought I'd add the example from the caddy documentation in the right syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,17 @@ localhost {
 }
 ```
 
+[HTTP Basic Authentication](https://caddyserver.com/docs/caddyfile/directives/basicauth) can be created like this:
+```
+caddy.basicauth: /secret/*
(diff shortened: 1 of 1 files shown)


darebfh:
Amazing! That's exactly what I was looking for - thank you, @nahakiole!

mykiwi (approved):

ozt:
I think it worth adding that when has is saved in docker-compose.yml, the 3 '$' signs need to be "escaped" with another dollar:
So:
 "password"
↓
$2a$14$WYgRwhWuywq/ln9nMrhwJunqGsKuSH0SNj1SDzqzHqYjuhqEtG0Py
↓
$$2a$$14$$WYgRwhWuywq/ln9nMrhwJunqGsKuSH0SNj1SDzqzHqYjuhqEtG0Py

lucaslorentz:
Added example to wiki page, let's use the wiki for example because I expect a lot of examples to be provided.
I will add a link to wiki in README soon

PR #452, opened 2023-01-22
Fixing custom images Dockerfile

Yoann-TYT:
Fixing custom images Dockerfile


francislavoie:
I don't think this is correct. See https://github.com/lucaslorentz/caddy-docker-proxy#new-module-name

Yoann-TYT:
Thanks a lot for your fast response :) 

You are right, just a timeout from github mislead me. 

I close it !

Issue #157, opened 2020-06-16
Controller will generate empty entries for containers it doesn't share a network with


rgdev:
If the controller don't share a Docker Network with the upstream containers / services that needs to be exposed through Caddy it prints a warning but still generates a block with no IPs causing 502s : 

```
caddytest.xxxxxxxx.xxx {
	reverse_proxy
}
```

For the sake of network isolation I'd rather not have the controller reachable from containers that don't absolutely requires it.

A way to make this possible could be to add a label on the upstream service (let's say caddy.network=my_net) and have caddy-docker-proxy resolve the container(s) IP(s) explicitly on this network. I believe that is how Traefik handles it internally.

This would allow this kind of network topology where the Caddy Controller is completely isolated from anything but the Caddy Servers : 

(simplified stack for readability)
```
services:
  caddy_server:
    ports:
      - 80:80
      - 443:443
    networks:
      - caddy_controller
      - caddy
    environment:
      - CADDY_DOCKER_MODE=server

  caddy_controller:
    networks:
      - caddy_controller
    environment:
      - CADDY_DOCKER_MODE=controller

  my_app:
    networks:
      - caddy
    deploy:
      labels:
      - caddy=myapp.domain.tld
      - caddy.network=caddy
      - caddy.reverse_proxy="{{upstreams 8000}}"
```

lucaslorentz:
Can you please try setting environment CADDY_DOCKER_VALIDATE_NETWORK=false in caddy-controller?

By default we do network validation, maybe we should change that default for controllers.

lucaslorentz:
Well, that doesn't fully solve your problem. Because if containers are in more than one network the wrong IP might be used.

I liked `caddy.network=my_net` idea, but I think we can define it only once in controller, instead of setting it on each service.

rgdev:
@lucaslorentz Adding CADDY_DOCKER_VALIDATE_NETWORK indeed solves the empty block issues but as you mentioned it ends up adding all the IPs attached to the container no matter if they are reachable by the Caddy Servers.

Defining the network label on the controller could be good for convenience in case you have a single network to bridge your Caddy Servers and all your exposed services however it would prevent adding more layers of network isolation. It might not be ideal to have many services all seeing each other on one network. I've seen clusters where each service had its own network and the front proxy was connected to each of those networks to keep things fully isolated.

[...]

PR #480, opened 2023-04-25
Per container ingress network override

kiwiz:
This PR adds support for overriding the ingress network for a container via a `caddy_ingress_network` label. Relates to #157

~Pushing this up for any early feedback - the code compiles, but I still need to test & add docs and tests are ready.~
Should be ready to go.

### DIFF
--- a/generator/containers.go
+++ b/generator/containers.go
@@ -17,8 +17,18 @@ func (g *CaddyfileGenerator) getContainerCaddyfile(container *types.Container, l
 func (g *CaddyfileGenerator) getContainerIPAddresses(container *types.Container, logger *zap.Logger, ingress bool) ([]string, error) {
 	ips := []string{}
 
-	for _, network := range container.NetworkSettings.Networks {
-		if !ingress || g.ingressNetworks[network.NetworkID] {
+	ingressNetworkFromLabel, overrideNetwork := container.Labels[IngressNetworkLabel]
+
+	for networkName, network := range container.NetworkSettings.Networks {
+		include := false
+
+		if overrideNetwork {
+			include = networkName == ingressNetworkFromLabel
+		} else {
+			include = !ingress || g.ingressNetworks[network.NetworkID]
+		}
+
+		if include {
 			ips = append(ips, network.IPAddress)
 		}
 	}
--- a/generator/containers.go
+++ b/generator/containers.go
@@ -14,7 +14,7 @@ func (g *CaddyfileGenerator) getContainerCaddyfile(container *types.Container, l
 	})
 }
 
-func (g *CaddyfileGenerator) getContainerIPAddresses(container *types.Container, logger *zap.Logger, ingress bool) ([]string, error) {
+func (g *CaddyfileGenerator) getContainerIPAddresses(container *types.Container, logger *zap.Logger, onlyIngressIps bool) ([]string, error) {
 	ips := []string{}
 
 	ingressNetworkFromLabel, overrideNetwork := container.Labels[IngressNetworkLabel]
@@ -22,10 +22,12 @@ func (g *CaddyfileGenerator) getContainerIPAddresses(container *types.Container,
 	for networkName, network := range container.NetworkSettings.Networks {
 		include := false
 
-		if overrideNetwork {
+		if !onlyIngressIps  {
+			include = true
+		} else if overrideNetwork {
 			include = networkName == ingressNetworkFromLabel
 		} else {
-			include = !ingress || g.ingressNetworks[network.NetworkID]
+			include = g.ingressNetworks[network.NetworkID]
 		}
 
 		if include {
--- a/generator/services.go
+++ b/generator/services.go
@@ -66,7 +66,7 @@ func (g *CaddyfileGenerator) getServiceTasksIps(service *swarm.Service, logger *
 		for _, task := range tasks {
 			if task.Status.State == swarm.TaskStateRunning {
 				hasRunningTasks = true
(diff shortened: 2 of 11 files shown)


lucaslorentz:
Hey, thanks @kiwiz for working on that!

The place you started extending is called only once when caddy is initialized and it will not update if new containers are created with new networks. It will also share the ingress network with all containers, instead of using it only for the container that declares the label.

My recommendation is to do it in https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/generator/containers.go#L17 and https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/generator/services.go#L52

### For containers.go

I did a docker inspect in a container and found the network name is in the key of the network map, as per the image below.
Screenshot 2023-04-25 at 19 45 02

I think this will hold true for docker, not sure if podman will be like that, but it should be fine for now.

So, I think changing this block:
```
	for _, network := range container.NetworkSettings.Networks {
		if !ingress || g.ingressNetworks[network.NetworkID] {
			ips = append(ips, network.IPAddress)
		}
	}
```

Into something like:
```
	ingressNetworkFromLabel := container.Labels[IngressNetworkLabel]
	for networkName, network := range container.NetworkSettings.Networks {
		if !ingress || g.ingressNetworks[network.NetworkID] || networkName == ingressNetworkFromLabel {
			ips = append(ips, network.IPAddress)
		}
	}
```
Might be enough, but I haven't tested.

### For services.go

I didn't sketch any code for this one, but I did inspect to check where the network name is:
Screenshot 2023-04-25 at 19 44 18

We could write similar code for this as well.

If you're not using docker swarm, feel free to implement it only for containers, I don't mind merging it without support for services for now.

kiwiz:
Thanks for the pointers! I updated the code and tested locally. Can confirm that the non-swarm side of things works. It's a bit different from your proposed solution, so feedback welcome.

I also implement similar logic on the Swarm side, but it's totally untested atm.

generator/containers.go
```diff
@@ -17,8 +17,18 @@ func (g *CaddyfileGenerator) getContainerCaddyfile(container *types.Container, l
 func (g *CaddyfileGenerator) getContainerIPAddresses(container *types.Container, logger *zap.Logger, ingress bool) ([]string, error) {
 	ips := []string{}
 
-	for _, network := range container.NetworkSettings.Networks {
-		if !ingress || g.ingressNetworks[network.NetworkID] {
+	ingressNetworkFromLabel, overrideNetwork := container.Labels[IngressNetworkLabel]
+
+	for networkName, network := range container.NetworkSettings.Networks {
+		include := false
```
kiwiz, line 23: There might be a simpler way to express this logic, but I believe the `caddy_ingress_network` label needs to be a complete override of the allowed networks. Otherwise, we could still return IPs from another network?

generator/services.go
```diff
@@ -66,8 +66,18 @@ func (g *CaddyfileGenerator) getServiceTasksIps(service *swarm.Service, logger *
 		for _, task := range tasks {
 			if task.Status.State == swarm.TaskStateRunning {
 				hasRunningTasks = true
+				ingressNetworkFromLabel, overrideNetwork := task.Labels[IngressNetworkLabel]
```
kiwiz, line 69: Might need to pull this label from `service` instead. IIRC, the task may have a different set of labels?

kiwiz: Yup. Updated to use `service`

kiwiz:
I've added Docker tests for swarm and standalone usage. Still working on the Go tests - unsure why they're failing atm.

kiwiz:
OK, tests fixed. However, I don't think `TestServices_OverrideIngressNetwork` is testing anything meaningful. Marking as ready for review.

lucaslorentz:
@kiwiz Looking very good!
Glad you added unit and integration tests.
I did one comment about "ingress" param, but that's all, once that's adjusted we can merge.

generator/services_test.go
```diff
@@ -138,6 +138,52 @@ func TestServices_ManualIngressNetwork(t *testing.T) {
...
+	}
+
+	const expectedCaddyfile = "service.testdomain.com {\n" +
+		"	reverse_proxy service\n" +
+		"}\n"
+
+	const expectedLogs = otherIngressNetworksMapLog + swarmIsAvailableLog
+
+	testGeneration(t, dockerClient, func(options *config.Options) {
+		options.IngressNetworks = []string{"other-network-name"}
```
lucaslorentz, line 183: You're right, this test is not meaningful as it is right now.
But swarm service proxying supports 2 modes, using dns name, or proxying to ips.
Add another option after this line like:
```
options.ProxyServiceTasks = true
```
This will swap the behaviour, and then adjust the `expectedCaddyfile` to match the IP addresses configured in `dockerClient.ServicesData`

generator/containers.go
kiwiz: Will go with `onlyIngressIps`

generator/services_test.go
kiwiz: Ah, I see. If I understand correctly, I should use `TestServiceTasks_ManualIngressNetwork` as a reference. Is overriding the ingress network to a virtual service something that should be tested?

lucaslorentz: No need

lucaslorentz:
Thanks @kiwiz!
Will do a new release with this

kiwiz:
:tada: Thanks!

lucaslorentz:
https://github.com/lucaslorentz/caddy-docker-proxy/releases/tag/v2.8.4
Build should push to docker hub soon

Issue #423, opened 2022-11-05
Support the `caddy run` flag `--envfile`


polarathene:
Presently using `--envfile` seems the best supported way to provide secrets like a DNS API token, without leaking into the container metadata (_which is visible from `docker inspect` on a host_), or to any other process within the container - both of which would happen with providing ENV to a container via `environment` and `envfile` in compose config.

Better would be the ability to source the secret file directly within the Caddyfile at runtime similar to `env` placeholder, but this is not a feature Caddy presently supports.

---

[Related issue raised at `caddy-dns/cloudflare`](https://github.com/caddy-dns/cloudflare/issues/46) (_requesting support to source a secret directly, instead of using it as an ENV_).

lucaslorentz:
Feature included in https://github.com/lucaslorentz/caddy-docker-proxy/releases/tag/v2.8.10

PR #548, opened 2023-11-26
Support `--envfile` CLI argument

thekondor:
Addresses #423:
- To stay consistent with `caddy`, the argument's name is `--envfile`.
- [godotenv library](https://github.com/lucaslorentz/caddy-docker-proxy/pull/548#issuecomment-1826739278) is used to parse & load envfile (was: ~Since `caddy` doesn't expose API for loading environment files, a corresponding code was borrowed with a copyright note.~)


loader.go
```diff
@@ -59,104 +60,114 @@ func logger() *zap.Logger {
...
-	if !dockerLoader.initialized {
-		dockerLoader.initialized = true
-		log := logger()
-
-		dockerClients := []docker.Client{}
-		for i, dockerSocket := range dockerLoader.options.DockerSockets {
-			// cf https://github.com/docker/go-docker/blob/master/client.go
-			// setenv to use NewEnvClient
-			// or manually
+	if dockerLoader.initialized {
```
thekondor, line 63: Since I touched this function, this `if` statement was inverted to flatten the implementation.

francislavoie:
Why is this necessary? You can load the env from file with Docker Compose already.

thekondor:
> Why is this necessary? You can load the env from file with Docker Compose already.

- #423 states exactly the use-case - to prevent leaking the secrets set through environment variables to metadata (`docker inspect`). `env_file` in a compose file exposes variables to a container's metadata anyway.
- `docker-compose` is not the only must-be-used-way to run `caddy-docker-proxy` (given that `docker-compose` is able to inject environment variables w/o publishing them to the metadata).

francislavoie:
https://github.com/joho/godotenv could be used instead of copying Caddy's impl. 

We originally wrote our own to keep things simple but it already grew a bit since. I'm considering switching to a lib at some point in the future

loader.go
```diff
@@ -59,104 +60,115 @@ func logger() *zap.Logger {
...
-				os.Setenv("DOCKER_CERT_PATH", dockerLoader.options.DockerCertsPath[i])
-			} else {
-				os.Unsetenv("DOCKER_CERT_PATH")
-			}
+	if envFile := dockerLoader.options.EnvFile; envFile != "" {
+		if err := godotenv.Load(dockerLoader.options.EnvFile); err != nil {
+			log.Error("Load variables from environment file failed", zap.Error(err), zap.String("envFile", dockerLoader.options.EnvFile))
+			return err
+		}
+		log.Info("environment file=%s loaded", zap.String("envFile", dockerLoader.options.EnvFile))
```
lucaslorentz, line 75: Can you please double-check what's in your logs? I wonder if %s is doing anything, given the envFile is passed as a structured field.
If %s is not being replaced, let's just remove `=%s`?

lucaslorentz: I double-checked, this is the output:
```
{"level":"info","ts":1701208881.6625867,"logger":"docker-proxy","msg":"environment file=%s loaded","envFile":"/etc/caddy/env"}
```
Can you please remove `=%s`?

lucaslorentz:
Code LGTM.
Requested a minor change to the log message, will give you some time to adjust it before merging.

lucaslorentz:
Merged, will do a new release with it so you can use it

PR #566, opened 2024-01-02
Update the whoami source 

dervomsee:
containous/whoami is deprecated and there is no arm image available. traefik/whoami should be used.
grafik

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,9 +646,9 @@ $ docker stack rm caddy-docker-demo
 ```
 $ docker run --name caddy -d -p 443:443 -v /var/run/docker.sock:/var/run/docker.sock lucaslorentz/caddy-docker-proxy:ci-alpine
 
-$ docker run --name whoami0 -d -l caddy=whoami0.example.com -l "caddy.reverse_proxy={{upstreams 80}}" -l caddy.tls=internal containous/whoami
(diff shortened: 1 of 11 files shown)


lucaslorentz:
Thanks for pointing that out!

2 thoughts about the change:
- I know Traefik is a great project, but it is a direct competitor to Caddy, we can easily avoid doing free advertising by just picking another image, like `hashicorp/http-echo`. It doesn't support arm/v7 though, only arm64.
- It would be nice to also update all other places that usees `containous/whoami`, like tests in `https://github.com/lucaslorentz/caddy-docker-proxy/tree/master/tests`. I understand if you're not willing to all that change, I can do it in a separate PR later.

@francislavoie any thoughts about using Traefik image?

francislavoie:
If the image is useful, there's no problem with using it IMO. It would be kinda petty to not use it just because of the org name 🤷‍♂️

lucaslorentz:
@francislavoie Sounds good.

@dervomsee Would you kindly update the other occurrences of `containous/whoami` in the readme as well?

dervomsee:
No problem, I can update the files tomorrow.
So traefik or a different project?

lucaslorentz:
Traefik is good.

dervomsee:
I updated the whoami source in all other files.

Issue #652, opened 2024-08-15
[Feature Request] Automatic local domain management


sinopsysHK:
Hello,

I really like this caddy-docker-proxy as it really helped me achieving what I wanted to do:

I have a bunch of web apps hosted on one server deployed via docker in my private lan.
Each of these web apps are therefore mapped to a different exposed ports to prevent conflicts.

So instead of requesting these web apps using http[s]://docker.mydomain:XYZ I'm happy to use caddy docker proxy to reach them through https://webappname.mydomain/
(I use a local acme server to dynamically generate certificates + my local dns to route *.mydomain to caddy.mydomain)

But I still have a little pebble in my shoe:  if ever I'm too lazy to type the full FQDN URL and I just want to query https://webappname/ caddy is not able to find related certificate and I land on my browser with an ERR_SSL_PROTOCOL_ERROR error.

The workaround is easy: just have to set both short and long addresses in the label: 

`--label caddy="webappname, webappname.mydomain"`

But with 25+ apps it becomes a bit cumbersome

So my request would it be possible to create an activable "local mode" via config to tell caddy docker proxy:

- if we want to activate this local mode or not
- what is the local domain name

And then allow a special pattern in the docker label to tell caddy docker proxy that this container is to be reversed-proxied within a local domain such as:
`--label caddy="webappname."`

And then when activated caddy docker proxy would automatically register the 2 reverse proxies:
- webappname
- webappname.\<localdomain\>
requesting one single certificate valid for "webappname,webappname.\<localdomain\>"

polarathene:
> I just want to query https://webappname/ caddy is not able to find related certificate and I land on my browser with an `ERR_SSL_PROTOCOL_ERROR` error.

That is a hostname, not a proper FQDN? 🤔 

You can just append `.localhost` for all FQDN to be routed to loopback (`localhost` / `127.0.0.1`) and have Caddy listen on that to respond to the address.

If the only intention is for a slightly shorter address for a service, then you should ask if typing is really the problem. https://gethomepage.dev is one of many services that help provide a single page to view your services for easy navigation. Many have a search / filter feature too, allowing you to type less.

Even so, with the longer name in the browser, if you have visited the URL before, it should autocomplete.

sinopsysHK:
Hello,

You're right https://webapp/ is alike hostname and not an FQDN which in my case would be https://webapp.compute.mydomain.here/

[...]

PR #655, opened 2024-08-19
Shortcut for hostnames in local domain

sinopsysHK:
Allow to label a container with

`--label caddy="@(myservice)"`

which will translate to reverse proxy to "_myservice_" and "_myservice_.<localdomain>" where local domain is provided as config parameter.

This is a very naive implementation as this is my first try of GO so do not hesitate to comment on the implementation and I'll try to improve.

I made an attempt to rather use GO template but this creates too much burden. 

Tentative implementation for #652

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,35 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            // working
+            "name": "Launch",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${fileDirname}",
+            //"logOutput": "dap",
+            //"showLog": true,
+            "stopOnEntry": true,
+            //"trace": "verbose",
+            "substitutePath": [{"from": "/home", "to": "/mnt/data1/home"}]
+        } ,
+       
+            {
+              "name": "Launch Caddy",
+              "type": "go",
+              "request": "launch",
+              "mode": "exec",
+              "program": "${workspaceFolder}/artifacts/binaries/linux/amd64/caddy", // the debug output of caddy
+              "args": ["docker-proxy", "--docker-sockets",  "\"unix:///var/run/docker.sock\"",
+                "--caddyfile-path", "~/dev/docker/scripts/mycaddy/Caddyfile",
+                ],
+                // --local-domain hkrose.ici --mode standalone --docker-sockets "unix:///var/run/docker.sock"
+              "cwd": "${workspaceFolder}"
+            }
+          
+    ]
+}
--- a/caddyfile/testdata/labels/local_domain.txt
+++ b/caddyfile/testdata/labels/local_domain.txt
@@ -1,6 +1,0 @@
-caddy               = {{ host app}}
-caddy.reverse_proxy = {{""}}
-----------
-app, app.localdomain {
-	reverse_proxy
-}
--- a/caddyfile/testdata/labels/local_domain.txt
+++ b/caddyfile/testdata/labels/local_domain.txt
@@ -1,0 +1,6 @@
+caddy               = {{ host app}}
+caddy.reverse_proxy = {{""}}
+----------
+app, app.localdomain {
+	reverse_proxy
+}
--- a/cmd.go
+++ b/cmd.go
@@ -258,6 +258,12 @@ func createOptions(flags caddycmd.Flags) *config.Options {
 		options.ScanStoppedContainers = scanStoppedContainersFlag
 	}
 
+	if localDomainEnv := os.Getenv("CADDY_DOCKER_LOCAL_DOMAIN"); localDomainEnv != "" {
+		options.LocalDomain = localDomainEnv
+	} else {
+		options.LocalDomain = localDomainFlag
+	}
+
 	if pollingIntervalEnv := os.Getenv("CADDY_DOCKER_POLLING_INTERVAL"); pollingIntervalEnv != "" {
 		if p, err := time.ParseDuration(pollingIntervalEnv); err != nil {
 			log.Error("Failed to parse CADDY_DOCKER_POLLING_INTERVAL", zap.String("CADDY_DOCKER_POLLING_INTERVAL", pollingIntervalEnv), zap.Error(err))
--- a/cmd.go
+++ b/cmd.go
@@ -153,6 +153,7 @@ func createOptions(flags caddycmd.Flags) *config.Options {
 	eventThrottleIntervalFlag := flags.Duration("event-throttle-interval")
 	modeFlag := flags.String("mode")
 	controllerSubnetFlag := flags.String("controller-network")
+	localDomainFlag := flags.String("local-domain")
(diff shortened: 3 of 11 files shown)


francislavoie:
I don't understand the motivation or goal here. I'm confused. This needs more explanation.

Also using syntax like this is risky because there's no guarantee Caddy may use the same kind of syntax in the future for something else. In fact, technically `@(whatever)` is a valid named matcher token right now (see https://caddyserver.com/docs/caddyfile/matchers#named-matchers)

sinopsysHK:
Hello,Motivation is to be able to easily reach all my web services hosted on my server with docker in a convenient way in my LAN within my private domain (no public exposure) by accessing them through a short and user friendly name rather than through a port number.And to make it even more convenient I want to be able to request my web apps without having to key in every time the full FQDN address but of course FQDN should work as well.I was relying so far on a url shortener to achieve something close to what I described but this workaround has limitations (confusing password safe that do not support properly urls with non standard ports).To achieve that with existing features of caddy (same issue with the alternate reverse proxies) I would have to register manually the 2 addresses for every web app I have such as:—label caddy=“webapp webapp.mydomain“Note: double registration is made mandatory once willing to use auto https (2 certificates are to be created).So I am looking for something more compact that would avoid to repeat things and reduce the dependencies between containers and the topology of my network.I’m happy to consider alternate syntax and agree that @() might be the most suitable if conflicting with existing one (sorry I am new adopter of caddy): any suggestion?On 19 Aug 2024, at 11:37 PM, Francis Lavoie ***@***.***> wrote:﻿
I don't understand the motivation or goal here. I'm confused. This needs more explanation.
Also using syntax like this is risky because there's no guarantee Caddy may use the same kind of syntax in the future for something else. In fact, technically @(whatever) is a valid named matcher token right now (see https://caddyserver.com/docs/caddyfile/matchers#named-matchers)

lucaslorentz:
I think the same can probably be achieved with CADDY env variables.
See https://caddyserver.com/docs/caddyfile-tutorial#environment-variables
Use label:
```
--label caddy="{$SITE_ADDRESS}"
```
You might need some escapes though.
And setting SITE_ADDRESS as an ENV variable in caddy container.

sinopsysHK:
well variable substitution will only achieve one part: decoupling a bit network topology with container but it will not avoid repeatingexport LOCAL_DOMAIN=myprivatedomaine.localSet for portainer:—label caddy=“portainer portainer.{$LOCAL_DOMAIN}”Set for Bitwarden:—label caddy=“bitwarden bitwarden.{$LOCAL_DOMAIN}”On 22 Aug 2024, at 2:40 AM, Lucas Lorentz ***@***.***> wrote:﻿
I think the same can probably be achieved with CADDY env variables.
See https://caddyserver.com/docs/caddyfile-tutorial#environment-variables
Use label:
--label caddy="{$SITE_ADDRESS}"

You might need some escapes though.
And setting SITE_ADDRESS as an ENV variable in caddy container.

lucaslorentz:
@sinopsysHK Thanks for the attempt to implement this.

Your current implementation is complex for such a niche feature. If we simplify it a bit, it is more interesting for the project to support it.

I agree with @francislavoie, I would rather not define a new special syntax like `@(whatever)`.

You mentioned problems using templates, could you please clarify what problems you had with it?

We already process go templates on all labels, it is already in place. Probably all you need to do is to add another function here: https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/generator/labels.go#L15

If you create a function called `createDomain`, that expects 1 string parameter that will be the domain prefix, then concatenate the prefix with the suffix defined in an ENV variable. You would be able to configure your containers with `--label 'caddy={{createDomain "myservice"}}'`

lucaslorentz:
Or, some even more flexible alternative would be to create a function that returns any env variable content, and use it as:
```
--label 'caddy=myservice.{{env \"LOCALDOMAIN\"}}'
```

sinopsysHK:
This is the first solution I tried and I felt notation was a bit too heavy because to set the expected prefix as a parameter of the expected function it has to be either a string literal either a variable.As a consequence prefix has to be surrounded with double quotes which might requires extra escaping when used with docker cli.All together something like—label caddy=‘{{createDomain “myservice”}}’or—label caddy=“{{createDomain \“myservice\”}}”appeared to be a bit wordy.But I understand that there is a trade off to find between lightness and consistency.Let me backtrack to the template based approach…On 23 Aug 2024, at 4:15 AM, Lucas Lorentz ***@***.***> wrote:﻿
@sinopsysHK Thanks for the attempt to implement this.
Your current implementation is complex for such a niche feature. If we simplify it a bit, it might be more interesting for the project to support it.
I agree with @francislavoie, I would rather not define a new special syntax like @(whatever).
You mentioned problems using templates, could you please clarify what problems you had with it?
We already process go templates on all labels, it is already in place. Probably all you need to do is to add another function here: https://github.com/lucaslorentz/caddy-docker-proxy/blob/master/generator/labels.go#L15
If you create a function called createDomain, that expects 1 string parameter that will be the domain prefix, then concatenate the prefix with the suffix defined in an ENV variable. You would be able to configure your containers with --label 'caddy={{createDomain "myservice"}}'

sinopsysHK:
Not sure to see how to achieve what you are proposing: how to get the prefix for duplication from outside the template scope!?Would you mind elaborating because this kind of notation looks more compactOn 23 Aug 2024, at 4:20 AM, Lucas Lorentz ***@***.***> wrote:﻿
Or, some even more flexible alternative would be to create a function that returns any env variable content, and use it as:
--label 'caddy=myservice.{{env \"LOCALDOMAIN\"}}'

sinopsysHK:
Ok I backtracked the implem to revert on the template based approach.

I used config to hold the Local Domain rather than an env variable to bring more control over it

polarathene:
> Motivation is to be able to easily reach all my web services hosted on my server with docker in a convenient way in my LAN within my private domain (no public exposure) by accessing them through a short and user friendly name rather than through a port number.
> 
> And to make it even more convenient I want to be able to request my web apps without having to key in every time the full FQDN address but of course FQDN should work as well.

This is [XY problem](https://en.wikipedia.org/wiki/XY_problem). Your interest is only in short URL for navigation reasons.

You instead want separate service like [Homepage](https://gethomepage.dev/latest/configs/docker/#automatic-service-discovery) (_links to Docker label service discovery feature_). That will provide you single page you visit with all your services available to navigate to.

Caddy / CDP will handle the actual routing from there and you don't have to think about ports in URL.

For any other needs with containers connecting to each other directly, you can use the Docker networks embedded DNS like you would with `reverse_proxy` directive in Caddy.

---

I don't understand the value this PR is trying to bring to CDP. It doesn't seem necessary or in high demand?

sinopsysHK:
If my motivation is a XY problem, your proposal is a workaround.

I came with a comprehensive user story and proposed an implementation for it.

Happy to rework if the implementation do not follow caddy coding standard or if convention isn't consistent with CDP.

Then your call to take this feature or not.

As commented in another thread CDP and caddy could be seen also as an alternate solution to suggested homepage which also avoids the need to recall each service port number.

polarathene:
For context, PR was closed as there are valid alternatives that can be used for the functionality they were seeking.

The alternative solutions are covered at the associated issue: https://github.com/lucaslorentz/caddy-docker-proxy/issues/652#issuecomment-2322727435

PR #743, opened 2025-08-27
Update Caddyfile on network event

tomleb:
I'm using https://git.sr.ht/~tomleb/docker-network-autoconnect to help manage networks between my applications (eg: bookstack, audiobookshelf, etc) and caddy. This means that the networks connected caddy are dynamically added/removed.

I found two problems with CDP, I fixed it with the commits here and that works for my use case.

1. CDP doesn't regenerates the Caddyfile when networks are connected/disconnected to caddy
2. The list of networks connected to caddy is computed once and reused.

This PR fixed both, by:
1. Reacting to docker network events. (This could be more specific eg: by filtering further the event to only caddy, but heh)
2. We're re-computing the list of networks every generation step

Not sure if this is needed or if this use case makes sense in general (probably though?), still I'm putting this PR up if anybody else needs this.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -54,13 +54,11 @@ func CreateGenerator(dockerClients []docker.Client, dockerUtils docker.Utils, op
 func (g *CaddyfileGenerator) GenerateCaddyfile(logger *zap.Logger) ([]byte, []string) {
 	var caddyfileBuffer bytes.Buffer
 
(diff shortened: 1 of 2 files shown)


lucaslorentz:
I wasn't aware of the use case for attaching new networks after the container is running.
Change is good, It will add a bit of overhead on every caddyfile update, but I think it is reasonable.

PR #778, opened 2026-03-03
feat: add Swarm mode with atomic Caddyfile rollout via configs

oneingan:
Implements the Swarm mode discussed in #773 (and the earlier atomic-write motivation from #766).

- Adds `--mode=swarm`: controller-only; renders a full Caddyfile from labels and rolls it out via immutable Swarm configs.
- Updates an existing Swarm service (`--swarm-service`) by swapping the mounted config at `/etc/caddy/Caddyfile` (or `--swarm-caddyfile-target`).
- Keeps Caddy workers socketless and the Admin API closed; rollout happens via `docker service update`, so disruption is governed by the service's update/rollback settings.
- Includes docs + `examples/swarm.yaml` and a swarm-mode test script.

Notes:
- Each config change creates a new Swarm config object named `<prefix>-<sha256>`; old configs are not garbage-collected yet (documented in README). A GC pass can be added as follow-up.
- This has been running on a production Swarm cluster for a couple of weeks without issues.

AI assistance disclosure: this PR was prepared with help from OpenCode (AI coding assistant) using OpenAI model `gpt-5.2-xhigh`.

### DIFF
--- a/loader.go
+++ b/loader.go
@@ -42,14 +42,19 @@ type DockerLoader struct {
 	lastVersion     int64
 	serversVersions *utils.StringInt64CMap
 	serversUpdating *utils.StringBoolCMap
+
+	// swarmServiceClientIndex is the docker client index used to manage the
+	// configured SwarmService. Only used when options.SwarmMode is enabled.
+	swarmServiceClientIndex int
 }
 
 // CreateDockerLoader creates a docker loader
 func CreateDockerLoader(options *config.Options) *DockerLoader {
 	return &DockerLoader{
-		options:         options,
-		serversVersions: utils.NewStringInt64CMap(),
-		serversUpdating: utils.NewStringBoolCMap(),
+		options:                 options,
+		serversVersions:         utils.NewStringInt64CMap(),
+		serversUpdating:         utils.NewStringBoolCMap(),
+		swarmServiceClientIndex: -1,
 	}
 }
 
@@ -159,6 +164,11 @@ func (dockerLoader *DockerLoader) Start() error {
 		zap.Strings("DockerCertsPath", dockerLoader.options.DockerCertsPath),
 		zap.Strings("DockerAPIsVersion", dockerLoader.options.DockerAPIsVersion),
 		zap.String("CaddyfileAutosavePath", CaddyfileAutosavePath),
+		zap.Bool("SwarmMode", dockerLoader.options.SwarmMode),
+		zap.String("SwarmService", dockerLoader.options.SwarmService),
+		zap.String("SwarmCaddyfileTarget", dockerLoader.options.SwarmCaddyfileTarget),
+		zap.String("SwarmConfigPrefix", dockerLoader.options.SwarmConfigPrefix),
+		zap.Int("SwarmConfigHashLen", dockerLoader.options.SwarmConfigHashLen),
 	)
 
 	ready := make(chan struct{})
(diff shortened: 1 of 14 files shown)


swarm_mode.go
```diff
@@ -0,0 +1,302 @@
...
+			Labels: map[string]string{},
+		},
+		Data: data,
+	}
+
+	resp, err := dockerClient.ConfigCreate(ctx, spec)
+	if err != nil {
+		// If another instance created it concurrently, inspect again.
+		if errdefs.IsConflict(err) {
+			return ensureSwarmConfig(ctx, dockerClient, name, data, fullHash)
```
lucaslorentz: This recursive call has no depth limit. If `ConfigCreate` keeps returning a conflict (e.g. a genuine name collision with different data), this will stack overflow.

Consider a bounded loop like the one in `ensureServiceCaddyfileConfig` just below (line 190).

loader.go
```diff
@@ -254,30 +264,40 @@ func (dockerLoader *DockerLoader) update() bool {
...
+
+			dockerLoader.lastJSONConfig = configJSON
+			dockerLoader.lastVersion++
 		}
+	}
 
-		if err != nil {
-			log.Error("Failed to convert caddyfile into json config", zap.Error(err))
+	if dockerLoader.options.SwarmMode {
+		if err := dockerLoader.updateSwarmService(); err != nil {
```
lucaslorentz: `updateSwarmService()` runs on every `update()` cycle, not just when `caddyfileChanged` is true. This means every polling interval / Docker event triggers `ServiceInspectWithRaw` + `ConfigList` even when nothing changed.

The existing flow only does expensive work (Adapt + server push) inside the `if caddyfileChanged` block. Could this be guarded the same way?

swarm_mode.go
```diff
@@ -0,0 +1,302 @@
...
+
+	"go.uber.org/zap"
+)
+
+const (
+	defaultSwarmCaddyfileTarget = "/etc/caddy/Caddyfile"
+	defaultSwarmConfigPrefix    = "caddyfile"
+	defaultSwarmConfigHashLen   = 32
+	maxSwarmConfigSizeBytes     = 1000 * 1024
+)
```
lucaslorentz: These defaults are also defined in the flag declarations in `cmd.go` (lines 75-86). In the existing code, defaults live only in the flag definitions. Having two sources means they can silently drift apart.

tests/swarm-mode/run.sh
```diff
@@ -0,0 +1,28 @@
...
+
+. ../functions.sh
+
+CONFIG_PREFIX="cdp-test-caddy_test"
+
+cleanup_configs() {
+  docker config ls --format "{{.Name}}" | grep "^${CONFIG_PREFIX}-" | xargs -r docker config rm >/dev/null || true
+}
+
+cleanup_configs
```
lucaslorentz: `cleanup_configs` runs here at the start but not on exit. `docker stack rm` (which the main test runner calls) won't remove standalone Swarm config objects created by this mode — unlike other tests, this one creates Docker objects outside the stack.

Consider adding a trap:
```bash
trap cleanup_configs EXIT
```

lucaslorentz:
Reviewed with Claude Opus

swarm_mode.go
lucaslorentz: Example:
```
fs.Int("swarm-config-hash-len", 32,
				"Length of sha256 hex used in generated Swarm config name (swarm mode only)")
```
Having a shared const would be nice

lucaslorentz:
A concern I have is the lack of config garbage collection. Each Caddyfile change creates a new Swarm config object (`<prefix>-<sha256>`) that is never removed. In a cluster with frequent deploys, this will silently accumulate and fill up Swarm's Raft store, which can degrade the entire cluster's performance and eventually prevent new configs/secrets from being created.

Would it be possible to tag the created configs with a label identifying them as managed by this controller, and include a cleanup routine that scans for old configs with that label and removes any that are no longer mounted on the target service?

Issue #734, opened 2025-07-24
can caddy docker proxy recognize labels on another server in any way


liamwibo:
hi i have 2 servers and i am trying to make caddy docker proxy see labels on the seconds server and reverse proxy to that second server.
is this possible in any way?

asieduernest12:
hmmm @liamwibo can you give a scenario and example like maybe a docker compose or two so someone can get a better understanding of what you are trying to achieve.

whn you say 2 servers right; do you mean a complete different 
server1_aws 
server2_azure
This, i dont know if its possible but look into docker swarm and overlay networking maybe

or you mean 
server1_vps_dockerComposeProjectA_nodeServer_withLabels_sharing_dockerProxyNetwork
server1_vps_dockerComposeProjectB_nodeServer_withLabels_sharing_dockerProxyNetwork
This is possible

liamwibo:
I want to achieve scenario 1 2 completely different nodes

asieduernest12:
Good and bad news, 
good new i got this behaviour  working with docker swarm running over 2 different nodes

Bad news: you have to go with docker swarm

Silver lining: I believe it is already mentioned in docker-proxy to use service labels when working with swarm, so basically what you want is supported within docker swarm

You can look over tthis sample project https://github.com/asieduernest12/SwarmDockerSwans

mcsdodo:
I'm on the same boat - for this purpose I've chained multiple caddy instances on my local network running on multiple docker compose hosts (portainer). Swarm is not an option for me for certain use-cases I'm implementing.

But what I would like to have is one Caddy reverse proxy and then client apps that would read the labels from containers and update master instance via API to reverse proxy for the services on those.

mcsdodo:
looking at the --mode flag - wouldn't be this behavior be achievable by combining multiple controllers with one server. Controllers would watch socket on their hosts, push changes to the server.

chrishoage:
> looking at the --mode flag - wouldn't be this behavior be achievable by combining multiple controllers with one server. Controllers would watch socket on their hosts, push changes to the server.

Looking at the code it relies on swarm APIs, so as you mention you can't use `docker stack deploy` and as such these services wouldn't be discoverable. 

I think that at least with the current state of this project there would need to be additional work to make the "controller" report to a KV store like redis, and have "sever" read from the redis for "service discovery" when not using `docker stack deploy` 

FWIW I have a similar use case in mind: 

[...]

PR #793, opened 2026-04-17
Make docker sockets optional with retry/reconnect

Rsgm:
## Summary
  - When `--docker-sockets` is configured with multiple sockets, each is now optional: a failed connection logs a warning and retries with exponential backoff instead of aborting startup.
  - Routes appear once a socket connects and are removed from the Caddyfile on the next regeneration if that socket drops.
  - New `--docker-retry-min` (default `1s`) and `--docker-retry-max` (default `30s`) control the backoff bounds. Available as `CADDY_DOCKER_RETRY_MIN` / `CADDY_DOCKER_RETRY_MAX`.
  - The single-socket default is unchanged — one unreachable socket is still a fatal startup error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,6 +500,8 @@ Run `caddy help docker-proxy` to see the raw flag output.
 | `--docker-sockets` | `CADDY_DOCKER_SOCKETS` | Comma-separated Docker sockets.<br>**Default:** `DOCKER_HOST` or the default socket |
 | `--docker-certs-path` | `CADDY_DOCKER_CERTS_PATH` | Comma-separated cert paths (one per socket; leave entry empty for sockets without certs) |
 | `--docker-apis-version` | `CADDY_DOCKER_APIS_VERSION` | Comma-separated API versions (one per socket) |
+| `--docker-retry-min` | `CADDY_DOCKER_RETRY_MIN` | Minimum retry interval for Docker socket reconnection (see [Optional Docker Sockets](#optional-docker-sockets)).<br>**Default:** `1s` |
+| `--docker-retry-max` | `CADDY_DOCKER_RETRY_MAX` | Maximum retry interval for Docker socket reconnection (see [Optional Docker Sockets](#optional-docker-sockets)).<br>**Default:** `30s` |
 | `--controller-network` | `CADDY_CONTROLLER_NETWORK` | Network allowed to configure the Caddy server, in CIDR (e.g. `10.200.200.0/24`) |
 | `--ingress-networks` | `CADDY_INGRESS_NETWORKS` | Comma-separated ingress networks connecting Caddy to containers.<br>**Default:** networks attached to the controller container |
 | `--caddyfile-path` | `CADDY_DOCKER_CADDYFILE_PATH` | Path to a base Caddyfile that will be extended with Docker sites |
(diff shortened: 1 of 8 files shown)


Rsgm:
This should also cover the use cases in #734.

lucaslorentz:
@Rsgm Ready for review? I see it is marked as draft

Rsgm:
Yes, this is ready.

I'm using this to build a multihost docker server with a central HTTP proxy. Each of the other docker hosts get a docker socket proxy exposed. This lets all this work without interrupting other services when hosts need to go down. Unfortunately, the best way to do the reverse proxying this way seems to be exposing ports on the host with labels pointing to the `host:port`, since they don't share docker networks. Docker swarm might be better for this, but network isolation was my main goal and I don't know if overlay networks allow that.

lucaslorentz:
## Review feedback — Claude (review requested by @lucaslorentz)

Overall direction is solid: the move from `[]docker.Client` to a `ClientProvider` closure plus per-socket lifecycle goroutines is the right shape, and keeping the default single-socket path strict preserves backwards compatibility well. The issues below are localized — none require redesigning the approach.

### 🔴 Blockers (correctness / resource bugs)

**1. Stale `swarmIsAvailable` indices when the connected-socket set shape changes** — `generator/generator.go` in `GenerateCaddyfile`
`getConnectedClients()` returns only non-nil clients, so the slice is dynamically reindexed by which sockets happen to be up. But `swarmIsAvailable[i]` is keyed by position in that reindexed slice and is only refreshed when the count changes or 60s pass.

Failure case: sockets `[A,B,C]` with `[A,C]` connected → `clients=[A,C]`, `swarmIsAvailable=[A_swarm, C_swarm]`. A drops and B reconnects in the same regen cycle → `clients=[B,C]`, count still 2, but `swarmIsAvailable[0]` still holds A's flag while representing B. For up to 60s, swarm configs/services are read from the wrong socket.

**Fix:** key swarm availability by socket identity, not slice index — either a `map[*socketManager]bool`, or move the flag onto `socketManager` and refresh on each successful connect.

**2. Docker client leak on every reconnect** — `socket_manager.go` `run()`
On disconnect, `setClient(sm, nil)` clears the reference but the underlying `*client.Client` is never `Close()`-d. The docker SDK holds an HTTP connection pool and idle goroutines behind that handle. In the flapping-daemon scenario this PR exists to support, every flap leaks.

**Fix:** call `Close()` on the old client before clearing. May require adding `Close()` to the `docker.Client` interface.

**3. Tight-loop risk on fast flap / `retryMin=0`** — `socket_manager.go` `run()`
After `listenEvents` returns, control falls straight back to `attemptConnect` with no sleep. If connect succeeds but events break immediately, the loop spins. Worse, `CADDY_DOCKER_RETRY_MIN=0` is accepted by the parser and `nextBackoff(0) == 0` → hot loop on any failing socket.

**Fix:** clamp `retryMin` to a sane minimum (e.g. 100ms) at parse time, and ensure at least one backoff sleep between iterations even on the \"successful connect → immediate error\" path.

**4. Data race on `sm.skipEvents`** — written from the events goroutine in `listenEvents`, cleared from `DockerLoader.update()` on the timer goroutine. No mutex or atomic. `go test -race` will flag it.
(The race existed before as `[]bool` on the loader, so not a severity regression — but this refactor is the right moment to close it out. `atomic.Bool` is a one-line change.)

### 🟢 Nice-to-have

- **`attemptConnect` has a misleading signature** — returns `(docker.Client, time.Duration)` but the duration is always `retryMin` and the single caller discards it. Simplify to `(docker.Client, error)`.
- **Log spam for long-down sockets** — `attemptConnect` emits `log.Warn` on every attempt (~2/min indefinitely for a decommissioned socket). Log first failure at Warn, subsequent at Debug until reconnect succeeds.
- **End-to-end test for `run()` lifecycle missing** — unit tests cover the pieces in isolation, but the reconnect lifecycle (connect → listen → disconnect → reconnect → trigger regen) isn't exercised. Add a test with a fake `connectFn` that fails N times then succeeds, plus an events channel that closes, asserting regen triggers on both edges.
- **Silent behavior change for existing multi-socket users** — pre-PR, `--docker-sockets=a,b` failed fast if either was unreachable; post-PR it retries forever. Please call this out in the PR description / release notes.

---

Happy to look at a follow-up revision. Nice refactor overall — the socket lifecycle abstraction is an improvement over the previous flow even before considering optional sockets.

— Claude (automated review via Claude Code, on behalf of @lucaslorentz)

lucaslorentz:
Also — @Rsgm, if helpful, I can open a follow-up PR against your branch with the four blockers fixed (swarm availability keyed by socket, client `Close()` on reconnect, backoff clamp + guaranteed sleep, `atomic.Bool` for `skipEvents`) plus the e2e test for the reconnect lifecycle. Just say the word and I'll send it your way.

— Claude

lucaslorentz:
> Docker swarm might be better for this, but network isolation was my main goal and I don't know if overlay networks allow that.

@Rsgm You can have multiple networks connecting your containers to CDP, and no network connecting your containers to each other.

PR #813, opened 2026-06-20
Migrate docker library to its successor, moby

shurane:
Clears the remaining few vulns that show up in `docker/docker`, which is now `moby/moby` as of v29: https://github.com/moby/moby

Let me know what you think. This is built on top of the other branch that bumps alpine from 3.20.3 to latest.


lucaslorentz:
That makes sense.
Thanks for working on it!
I think I rather pinning an alpine version though.
I'm merging and will adjust that afterwards and configure dependabot to update alpine for us.

shurane:
Yup, pinning an alpine version + having dependabot update the versions also makes sense. This was way quicker turnaround than I expected, thanks!

Issue #804, opened 2026-05-24
CADDY_ADMIN


llitz:
is there a chance that someting is still broken in CADDY_ADMIN with unix sockets?

I have it set to unix//run/caddy-admin.sock and it seems to load up properly:

```
caddy-1         | {"level":"info","ts":1779595454.096892,"logger":"admin","msg":"admin endpoint started","address":"unix//run/caddy-admin.sock","enforce_origin":false,"origins":[]}
caddy-1         | {"level":"warn","ts":1779595454.0969188,"logger":"admin","msg":"admin endpoint on open interface; host checking disabled","address":"unix//run/caddy-admin.sock"}
```

but when it actually try to apply the config it fails trying to connect to localhost:2019 - a non existent port

```
caddy-1         | {"level":"info","ts":1779595484.1135478,"logger":"docker-proxy","msg":"Sending configuration to","server":"localhost"}
caddy-1         | {"level":"error","ts":1779595484.1143422,"logger":"docker-proxy","msg":"Failed to send configuration to","server":"localhost","error":"Post \"http://localhost:2019/load\": dial tcp [::1]:2019: connect: connection refused"}
```

this is running in standalone mode

Idk if this is supposed to reach that part of the code, but line 316 on loader.go forces the connection to be http and I am not finding where it should've used/detected the unix path (which I assume is failing.

am I missing something?

lucaslorentz:
Should have been fixed by https://github.com/lucaslorentz/caddy-docker-proxy/pull/817
Please try again and let us know if you see any issues.

PR #816, opened 2026-06-22
Fix admin reloads over Unix sockets (#1)

boblangley:
Related to lucaslorentz/caddy-docker-proxy#804

## Summary

Fixes admin reloads when `CADDY_ADMIN` is configured to use a Caddy Unix socket address, such as `unix//run/caddy-admin.sock`.

The loader now builds the admin API request target from the configured admin listen address when one is provided. Unix socket admin endpoints use a custom HTTP transport that dials the socket directly, while the request URL uses `127.0.0.1` only to satisfy `net/http` URL requirements. TCP behavior remains unchanged for the default path and now honors explicit TCP admin ports.

## Points of interest

- `loader.go:316-343`: `updateServer` now uses the resolved admin API endpoint, closes response bodies, and runs endpoint cleanup when needed.
- `loader.go:361-392`: `adminAPIEndpoint` handles default TCP, explicit TCP, wildcard TCP bind hosts, and Unix socket admin listeners.
- `loader_test.go:21-109`: Adds coverage for default TCP, explicit TCP ports, Unix socket requests, invalid listen addresses, and port range rejection.

### D1 Preserve the default TCP reload path

When `AdminListen` is unset, the loader still returns `http://<server>:2019/load` directly instead of parsing the synthesized Caddy listen address.

**Alternatives considered**

- D1A: Always parse `tcp/<server>:2019`. Not selected because it could turn previously tolerated server values into parse failures.

**Assumptions used**

- The default path should remain behaviorally compatible with the pre-existing implementation.

### D2 Honor explicit TCP admin ports while replacing wildcard hosts

When `AdminListen` is an explicit TCP address, the reload URL now uses the configured port. If the configured host is a wildcard bind address, such as `0.0.0.0`, the loader dials the controlled server address instead.

**Alternatives considered**

- D2A: Dial the parsed TCP host directly. Not selected because dialing `0.0.0.0` is the wrong behavior for controlled remote servers.
- D2B: Keep TCP hard-coded to `:2019`. Not selected because explicit TCP admin ports should work consistently with explicit Unix socket admin addresses.

**Assumptions used**

- Wildcard listen hosts describe where Caddy binds, not the routable address the loader should dial.

### DIFF
--- a/loader.go
+++ b/loader.go
@@ -6,8 +6,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"path/filepath"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -310,7 +313,14 @@ func (dockerLoader *DockerLoader) updateServer(wg *sync.WaitGroup, server string
 	log := logger()
 	log.Info("Sending configuration to", zap.String("server", server))
 
-	url := "http://" + server + ":2019/load"
+	client, url, cleanup, err := dockerLoader.adminAPIEndpoint(server)
+	if err != nil {
+		log.Error("Failed to determine admin API endpoint for", zap.String("server", server), zap.Error(err))
+		return
+	}
+	if cleanup != nil {
+		defer cleanup()
+	}
 
 	postBody, err := dockerLoader.prepareServerConfig(server)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lucaslorentz:
Hey @boblangley , I'm thinking about doing pushes of caddyfile to local server without Admin API.
In the past I've decided to push to local server with Admin API to have a single way of pushing for both local and remote servers. But it clearly has disadvantages, like the issue you're fixing and the fact it requires admin API to be on.

Do you think https://github.com/lucaslorentz/caddy-docker-proxy/pull/817 would fix your issue as well?

boblangley:
Yes, for my use case that would fix it, and indeed is the preferred approach.

I hesitated to suggest the same change because I was trying to conform to the approach you applied and suggest the minimal change.

PR #819, opened 2026-06-26
fix: ensure local load keeps admin disabled

boblangley:
## Summary

This builds on the local in-process reload path by ensuring it does not accidentally reopen Caddy’s admin endpoint.

For localhost reloads, `caddy-docker-proxy` no longer needs to POST to Caddy’s admin API because the config is applied directly with `caddy.Load`. The remote-server reload path still needs an admin endpoint, but the local path does not.

## Bug

Before this change, `prepareServerConfig("localhost")` still applied the same fallback behavior used for remote controlled servers:

- if the generated config had no `admin` block, it injected `admin.listen`
- if the generated config had `admin off`, it replaced that with `admin.listen`
- if `AdminListen` was configured, it used that value
- otherwise it fell back to `tcp/localhost:2019`

That made sense when localhost reloads also depended on the admin API. After switching localhost reloads to `caddy.Load`, this fallback undermines the reason to use the `caddy.Load` API instead of the admin API.

The practical effect was that a process started with admin disabled could have admin silently re-enabled on the first local config load. In standalone/local mode, the loader would prepare a config with an injected admin listener, then `caddy.Load` would apply that config and Caddy would start the admin endpoint.

## Fix

The local and remote paths now diverge:

- `localhost` loads config in-process and keeps admin disabled when the generated config had admin disabled or omitted
- remote controlled servers still receive the admin listener fallback because the controller still pushes config to them through the admin API
- explicit admin configuration is still respected

## Test Coverage

The existing local-load integration test now also verifies that the configured admin address does not become reachable after `updateServer("localhost")`.

The first test commit demonstrates the failure against the current implementation: Caddy starts the admin endpoint and responds to `/config/`.

The fix commit makes the same test pass by keeping admin disabled for local in-process loads.

### DIFF
--- a/loader.go
+++ b/loader.go
@@ -349,9 +349,14 @@ func (dockerLoader *DockerLoader) prepareServerConfig(server string) ([]byte, er
 		return nil, err
 	}
 
-	// Respect an explicit admin endpoint, but override "admin off" since the
-	// plugin requires the admin API.
-	if config.Admin == nil || config.Admin.Disabled {
+	// Local loads happen in-process, so keep the admin API disabled unless the
+	// config explicitly enables it. Remote servers still need an admin endpoint
+	// for controller pushes.
+	if server == "localhost" {
+		if config.Admin == nil || config.Admin.Disabled {
+			config.Admin = &caddy.AdminConfig{Disabled: true}
+		}
+	} else if config.Admin == nil || config.Admin.Disabled {
 		config.Admin = &caddy.AdminConfig{Listen: getServerAdminListen(dockerLoader.options, server)}
 	}
 
--- a/loader.go
+++ b/loader.go
@@ -352,12 +352,12 @@ func (dockerLoader *DockerLoader) prepareServerConfig(server string) ([]byte, er
 	// Local loads happen in-process, so keep the admin API disabled unless the
 	// config explicitly enables it. Remote servers still need an admin endpoint
 	// for controller pushes.
-	if server == "localhost" {
-		if config.Admin == nil || config.Admin.Disabled {
+	if config.Admin == nil || config.Admin.Disabled {
+		if server == localServer {
 			config.Admin = &caddy.AdminConfig{Disabled: true}
+		} else {
+			config.Admin = &caddy.AdminConfig{Listen: getServerAdminListen(dockerLoader.options, server)}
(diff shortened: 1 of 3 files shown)


loader.go
```diff
@@ -349,17 +384,22 @@ func (dockerLoader *DockerLoader) prepareServerConfig(server string) ([]byte, er
...
 	}
 
-	// Respect an explicit admin endpoint, but override "admin off" since the
-	// plugin requires the admin API.
-	if config.Admin == nil || config.Admin.Disabled {
+	// Local loads happen in-process, so keep the admin API disabled unless the
+	// config explicitly enables it. Remote servers still need an admin endpoint
+	// for controller pushes.
+	if server == "localhost" {
+		if config.Admin == nil || config.Admin.Disabled {
```
lucaslorentz: I wonder what happens if admin is set and we do a local load?
Would it delete the admin config that was previously set?

lucaslorentz (commented):
AI Review:

Correct fix — this is the logical follow-up to the local-load refactor, and the comment in `push.go` already states local push must not depend on admin being enabled. Nice catch that `Admin == nil` also needs explicit `Disabled: true`, since Caddy otherwise defaults admin on at `localhost:2019`. Tests look good.

Two small things:

1. Use the existing `localServer` const instead of the `"localhost"` literal (defined at loader.go:298, used at :362).

2. Avoid duplicating the `Admin == nil || Admin.Disabled` predicate:

```go
if config.Admin == nil || config.Admin.Disabled {
    if server == localServer {
        config.Admin = &caddy.AdminConfig{Disabled: true}
    } else {
        config.Admin = &caddy.AdminConfig{Listen: getServerAdminListen(dockerLoader.options, server)}
    }
}
```


---
_Reviewed by Claude (Claude Code), on behalf of @lucaslorentz._

loader.go
lucaslorentz: Never mind, now I remember this better.
This configures the admin during push of a config in case it is not configured yet.

lucaslorentz:
Will give you some time to do the minor adjustments above, but I'm happy to land without them as well. Thanks for the improvement.

boblangley:
I am happy to make those adjustments.

PR #821, opened 2026-07-01
fix: skip container IP addresses for non-running containers

DVDAndroid:
After the latest update (v2.13.0), new warnings started appearing in logs that were not present before:

```json
{"level":"warn","ts":1782899007.4949086,"logger":"docker-proxy","msg":"Container is not in same network as caddy","container":"firefox","container networks":["network1","network2"],"ingress networks":["network2"]}
```
and `CADDY_INGRESS_NETWORKS=network2`

The firefox container in this case is stopped.

I've set `CADDY_DOCKER_SCAN_STOPPED_CONTAINERS=true` in order to support Sablier, so stopped containers are intentionally included in the scan and can be started on demand.

With stopped containers included in the scan, the network check is executed on containers that have no active runtime networking state. This leads to false-positive warnings even though the setup is valid.

### Fix

Skip IP resolution and network validation when the container is not running.

### Tests

Added and updated some tests logic + manual testing with custom caddy image

### Reference

This probably happened due to #813, in particular https://github.com/lucaslorentz/caddy-docker-proxy/commit/21c56d57c119d833a9e9ec4f53496613376aba75#diff-6e8b88a8d296f10ad00b020c06d86a1f6647223c712c1a3e95f0701d1d759881R36

### DIFF
--- a/docker/client_mock.go
+++ b/docker/client_mock.go
@@ -3,7 +3,7 @@ package docker
 import (
 	"context"
 
-	"github.com/moby/moby/api/types/container"
+	mobyContainer "github.com/moby/moby/api/types/container"
 	"github.com/moby/moby/api/types/events"
 	"github.com/moby/moby/api/types/network"
 	"github.com/moby/moby/api/types/swarm"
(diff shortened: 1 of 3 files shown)


lucaslorentz:
@DVDAndroid thanks for the fix.
I want to try this different approach though: https://github.com/lucaslorentz/caddy-docker-proxy/pull/822
Where we would still warn if we detect containers that don't share any network with caddy server, even stopped containers.
If it doesn't work well we can restore your PR and skip all validations for stopped containers.

PR #826, opened 2026-07-21
Add read-only docker-socket-proxy example and security note

spilbros:
## What
- Adds `examples/socket-proxy.yaml`: a quickstart variant where Caddy reaches
  the Docker API through docker-socket-proxy (read-only, POST=0) instead of
  mounting `/var/run/docker.sock`.
- Adds a short security note next to the socket mount in the README.

## Why
The quickstart and existing examples mount the Docker socket directly. That's
fine to get started, but it grants the Caddy container root-equivalent access
to the host, while caddy-docker-proxy only ever needs read access to
container/network metadata and the event stream. Routing that through a
read-only socket proxy is a small, well-known hardening step that the docs
don't currently mention as an option.

I run this exact config in production, so the example is tested rather than
theoretical.

Happy to drop the README change and keep this example-only if you'd prefer.


lucaslorentz:
The pull request looks solid. But I'm not fully convinced we should push users towards doing that to be honest.

Can you help me understand one bit about the security of the solution? Why is letting https://github.com/Tecnativa/docker-socket-proxy have full access to the docker socket different and better than having caddy docker proxy have full access to the docker socket?

